Issue #5, opened 2018-07-09
pipeline for storm track


aappling-usgs:
at a minimum, we need the location of the center of the storm as a function of time, possibly interpolated to match the final timesteps (see #1).

if we want to be able to support non-hurricane flood vizzies, acquiring the storm track should be optional.

also consider a fancier representation of the storm as a dot of changing size or color, or a cone/polygon of [observed] influence. this would involve finding and acquiring more data from somewhere to inform the shape/size dimensions.

aappling-usgs:
split out the remaining (nice-to-have) tasks into #51 and #49, which makes this issue now closeable.

PR #16, opened 2018-07-09
initial storm track fetching

aappling-usgs:
First step in #5 pipeline for storm track: download the .zip file from the national hurricane center.

Also got some project stuff going:

* viz_config.yml is the YAML file where we can put all our configuration parameters that will distinguish one storm from another

* I added an example of how I think we should use that viz_config: it's loaded for use by all files in lib.yml, so every remake recipe can assume the existence of `viz_config`. But there should be a lot of subsetting recipes, like this:
    ```yml
      storm_track_cfg:
        command: viz_config[I('storm_code')]
    ```
    where `storm_track_cfg` then gets passed into specific functions. If we don't use subsetting recipes, then anytime anything changes in the config, _everything_ will be considered out of date and will be rebuilt by remake/scipiper. That's no good, so these subsetting functions will help us only rerun what needs to be rerun. I predict that having function-specific cfg subsets, like `storm_track_cfg` going into `fetch_storm_track()`, will be convenient in case some viz_config elements are required for multiple subsets.

* I replumbed the 1_lib.yml, 1_fetch.yml, 2_process.yml, 6_visualize.yml, and remake.yml files just a bit so that each depends on the next - was needed to fix redundant `include`s from before this PR. The way I did this makes a strong coupling among phases, which I think should be OK/good for this relatively quick analysis pipeline. But we could opt for a decoupled approach if needed...

### DIFF
--- a/viz_config.yml
+++ b/viz_config.yml
@@ -1,0 +1,1 @@
+storm_code: al152017 # find the right code at https://www.nhc.noaa.gov/gis/archive_besttrack.php?year=2018
--- a/.gitignore
+++ b/.gitignore
@@ -2,3 +2,7 @@
 .Rhistory
 .RData
 .Ruserdata
+.httr-oauth
+.remake
+*.zip
+*.rds
--- a/1_fetch.yml
+++ b/1_fetch.yml
@@ -4,14 +4,23 @@ include:
   - lib.yml
 
 packages:
+  - httr
 
 file_extensions:
-  - feather
   - ind
 
 sources:
+  - 1_fetch/src/fetch_storm_track.R
 
 targets:
 
   1_fetch:
     depends:
+      - 1_fetch/out/storm_track.zip
+
+  storm_track_cfg:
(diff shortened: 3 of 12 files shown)


viz_config.yml
```diff
@@ -0,0 +1 @@
+storm_code: al152017 # find the right code at https://www.nhc.noaa.gov/gis/archive_besttrack.php?year=2018
```
jread-usgs: Looks fine to me. In the past we did split this as "ocean" "stormnum" and "year" (or something like that). I don't really see any reason we need to split them...just noting the difference and this seems :+1: to me.

aappling-usgs: yeah, i saw that and decided to merge the three parameters because i also couldn't see a big need to split them. it'd be one thing if the user already knew the answer to all three params independently and just had to pass them in to get the correctly formatted code, but because somebody will always have to go to the NHC website to find out which storm number it is, i think it makes sense to copy-paste the entire 8-character code at that point.

PR #17, opened 2018-07-09
Get streamflow

jzwart:
Initial start of function for fetching streamflow data based on site_ids and dates. I imagine these sites will be stored in the `1_fetch/out/` folder as a .tsv or yaml  - I just stored stored a temporary one as a yaml for now. 

`dates.yml` is also temporary right now and will come from the storm config


1_fetch/cfg/dates.yml
```diff
@@ -0,0 +1,2 @@
+start: 2017-05-01
+end: 2017-05-10
```
aappling-usgs: departing from our practice in other projects, I'd like to try getting all the config info that's unique to a storm event into a single file (viz_config.yml in the top-level folder). that would include these dates. goal is to make it really fast to spin up a viz for a new storm, which i think would be facilitated by having all the info you need to edit in just one file.

1_fetch/cfg/stream_params.yml
```diff
@@ -0,0 +1,3 @@
+flow:
+  - '00060'
```
aappling-usgs: this config info is great right where it is.

1_fetch/cfg/dates.yml
aappling-usgs: going with jordan's suggestion in slack, our Harvey viz start/end dates were "2017-08-25" to "2017-09-02"

1_fetch.yml
```diff
@@ -4,18 +4,21 @@ include:
   - lib.yml
 
 packages:
+  - dataRetrieval
   - httr
```
aappling-usgs: i think we'll need `yaml` in here. i originally hoped that packages listed in lib.yml would be available to all inheriting yml files, but i think i learned that's not the case.

1_fetch.yml
```diff
@@ -4,18 +4,21 @@ include:
...
 sources:
+  - 1_fetch/src/fetch_streamflow.R
   - 1_fetch/src/fetch_storm_track.R
 
 targets:
 
   1_fetch:
     depends:
+      - 1_fetch/out/streamflow.rds.ind
       - 1_fetch/out/storm_track.zip
```
aappling-usgs: oops, good call - the storm_track depends item should also point to the .ind version of that file.

1_fetch/out/subset_sites.yml
```diff
@@ -0,0 +1,2 @@
+site_id:
+  - '07374000'
```
aappling-usgs: we can wait on this one until we know how we're handling sites, but the site list should also go into the shared viz_config.yml eventually

1_fetch/src/fetch_streamflow.R
```diff
@@ -0,0 +1,15 @@
+fetch_streamflow <- function(ind_file, sites_yml, dates_yml, stream_params_yml, gd_config) {
+
+  dates <- yaml::yaml.load_file(dates_yml) # these dates are user input right now but should be dynamic based on storm
+  pcodes <- yaml::yaml.load_file(stream_params_yml)
+  site_ids <- yaml::yaml.load_file(sites_yml)
```
lindsaycarr: I think we should pass these in as the already loaded yamls. Then we won't have to do this every time. Something like: https://github.com/USGS-R/nawqa_wqp/blob/master/1_wqpdata.yml#L30-L36

1_fetch/out/subset_sites.yml
lindsaycarr: ^ that seems like it would be difficult to do since the `viz_config` is editable by hand, but the site list will be auto-generated. Is it a good idea to mix that within the same file?

aappling-usgs: i was under the impression that we had hand-picked final sites for at least one vizzy. is that not true?

lindsaycarr: Hmm I was under the impression we were auto-generating this list as much as possible. Let's bring this over to slack

1_fetch/cfg/dates.yml
jzwart: I'm moving to viz_config on next push with some other changes

PR #18, opened 2018-07-09
getting this up for blodg

jread-usgs:
I need to do some things to this work (including documentation) but I think I am a blocker for others, so getting this up quick

### DIFF
--- a/1_fetch/src/map_utils.R
+++ b/1_fetch/src/map_utils.R
@@ -1,0 +1,61 @@
+fetch.map_data <- function(viz) {
+  deps <- readDepends(viz)
+  checkRequired(deps, "viewbox_limits")
+  viewbox <- deps[["viewbox_limits"]]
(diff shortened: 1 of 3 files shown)


1_fetch/src/map_utils.R
```diff
@@ -0,0 +1,40 @@
...
+    sf::st_transform(bbox_poly, return_crs)
+  }
+  else {
+    bbox_poly
+  }
+}
+
+
+
+as_view_polygon <- function(object) {
```
aappling-usgs: arg name could be a little more specific...maybe `view_config`?

jread-usgs: Since we are giving it an object that is a list that could have lots of different fields, I thought this pattern was closer to `viz` that other names that would be more specific.

aappling-usgs: But the object must have fields for `bbox`, `projection`, and `aspect`, right? And we ignore all others? That makes it pretty much view-specific and bbox-specific IMO.

PR #19, opened 2018-07-09
fixing a few errors in storm track fetch

aappling-usgs:
Oops, oops, and oops again.

Main target for storm_track fetch should have been the .ind file.

Should have been downloading the zip file to the data file, not the ind file.

Somehow failed to post an up-to-date build file earlier.


jread-usgs:
Should've caught that when I reviewed. Sorry, didn't build it.

PR #22, opened 2018-07-09
Fetch part of precip pipeline.

dblodgett-usgs:
Need to tackle the data access still, but this should be what we need for the spatial part.

Basically doing this:
screen shot 2018-07-09 at 2 52 55 pm


dblodgett-usgs:
I'm just going to tag the values part of this on too. 

I built this for a little zoomed in bounding box to verify that it works... seems pretty decent. 
screen shot 2018-07-09 at 8 22 11 pm
screen shot 2018-07-09 at 8 28 23 pm

This PR comes along with two .rds files that are in drive. One that is an intermediate spatial file -- it could be eliminated in hindsight. The main precip data.frame will be a _very_ long three column data.frame with the polygon geometry repeated for ease of plotting. We could change this up so there is one geometry .rds that has an ID column and one data .rds that has the ID to join up to the geometry. Thoughts?

screen shot 2018-07-09 at 8 32 05 pm
screen shot 2018-07-09 at 8 29 34 pm

jread-usgs:
This is looking good! reviewing now, but thank you for the details and images. Very helpful.

1_fetch.yml
```diff
@@ -67,3 +69,8 @@ targets:
   focus_geoms:
     command: fetch_geoms(focus_geoms_config, within = view_polygon)
 
+  1_fetch/out/precip_spatial.rds.ind:
+    command: fetch_precip_spatial(ind_file=target_name,
+                                view_config = view_config)
```
jread-usgs: I think you want `view_polygon`, not `view_config`

1_fetch/src/fetch_precip_spatial.R
```diff
@@ -0,0 +1,51 @@
...
+                          y = matrix(y, ncol = 1),
+                          lon = matrix(lon, ncol = 1),
+                          lat = matrix(lat, ncol = 1))
+
+  sf_points <- sf::st_as_sf(sf_points,
+                            coords = c("lon", "lat"),
+                            crs = "+init=epsg:4326",
+                            agr = "constant")
+
+  bbox <- bbox_to_polygon(bbox = view_config$bbox,
```
jread-usgs: This looks good, but I think you should use `view_polygon` as that gives the whole visible domain, vs the initial bbox isn't filled in after the projection and aspect ratio are applied.

1_fetch.yml
dblodgett-usgs: 👍 will give that a go -- don't think it existed when I started.

1_fetch/src/fetch_precip_spatial.R
dblodgett-usgs: Output looks something like this now: (code to generate this is pasted below.)
screen shot 2018-07-09 at 9 31 59 pm

Note that I am now creating a reusable "precip_spatial" and a non-spatial "precip" .rds file. as shown below. -- the NaNs are outside the domain of the data.

screen shot 2018-07-09 at 9 16 37 pm

Side now, this number of polygons renders _very_ slowly. I think we are going to want to do some resampling to a raster or visualization grid of some sort.

```r
precip_spatial <- readRDS("1_fetch/out/precip_spatial.rds")
precip <- readRDS("1_fetch/out/precip_values.rds")

view_config <- scmake("view_config")
source("1_fetch/src/map_utils.R")

bbox <- bbox_to_polygon(view_config$bbox, 
                        return_crs = view_config$projection)
county <- fetch_geoms(scmake("focus_geoms_config"), 
                      crs = view_config$projection, 
                      within = bbox)

fil <- precip$time == precip$time[1]

p <- left_join(select(precip_spatial, id), 
               select(filter(precip, fil), -time), 
               by = "id")

plot(bbox, lwd = 3, main = precip$time[1])
plot(p["precip"], border = "transparent", add = TRUE)
plot(county$geometry, add = TRUE)
plot(bbox, lwd = 3, add = TRUE)
```

dblodgett-usgs:
I think this is ready for merge unless there are changes you want to see in the latest check in @jread-usgs.

1_fetch/src/fetch_precip_values.R
```diff
@@ -0,0 +1,77 @@
...
+    precip[,,t_counter] <-
+      ncdf4::ncvar_get(nc, nc$var$Total_precipitation_surface_1_Hour_Accumulation,
+                       start <- c(min(x_inds),
+                                  min(y_inds),
+                                  step)[dimid_order],
+                       count <- c(length(x_inds),
+                                  length(y_inds),
+                                  1)[dimid_order])
+    t_counter <- t_counter + 1
+    message(paste(t_counter, length(t_inds)))
```
jread-usgs: Not worth changing, but the last `message` is `192 191`. Next time maybe do `1 out of 191` and put the `message()` above the counter increment. 

and if you want to give a progress progression, 
```r
if (t_counter %% 10 == 0){
   message(paste(t_counter, 'of', length(t_inds)))
}
```
Will print fewer lines (every ten)

1_fetch/src/fetch_precip_values.R
```diff
@@ -0,0 +1,77 @@
...
+  precip <- precip %>%
+    left_join(x_inds, by = "d1") %>%
+    left_join(y_inds, by = "d2") %>%
+    left_join(t_inds, by = "d3") %>%
+    inner_join(sf::st_set_geometry(precip_spatial, NULL),
+               by = c("x", "y")) %>%
+    left_join(data.frame(time = t_vals,
+                         t = seq_len(length(t_vals))),
+              by = "t") %>%
+    select(precip, time, id)
```
jread-usgs: Seems fine, but any reason you used a skinny data.frame instead of a 3d matrix of values? Having not played around w/ either, not sure if this one offers advantages that I am not seeing. Given all of the repeated values in the `$time` field (191 unique values in 37880266 rows), I think the data.frame format is quite a bit bigger.

1_fetch/src/fetch_precip_values.R
```diff
@@ -0,0 +1,77 @@
...
+  time_step<-time_units[1]
+  date_origin<-time_units[3]
+
+  if(grepl("hour", time_step, ignore.case = TRUE)) {
+    multiplier <- (60^2)
+  } else {
+    stop("only hour time steps supported so far")
+  }
+
+  as.POSIXct(t_dim$vals * multiplier,
```
jread-usgs: This might be a lot more readable w/ piping `%>%`

1_fetch/src/fetch_precip_spatial.R
```diff
@@ -0,0 +1,50 @@
...
+
+  sf_points <- sf::st_as_sf(sf_points,
+                            coords = c("lon", "lat"),
+                            crs = "+init=epsg:4326",
+                            agr = "constant")
+
+  sf_points_sub <- sf::st_intersection(
+    sf::st_transform(sf_points,
+                     sf::st_crs(view_polygon)),
+    view_polygon)
```
jread-usgs: Hmm, even w/ `view_polygon`, I seem to be seeing only a subset of the domain: 
image

jread-usgs:
with 
```r
precip_spatial <- readRDS("1_fetch/out/precip_spatial.rds")
precip <- readRDS("1_fetch/out/precip_values.rds")

view_polygon <- scmake("view_polygon")
source("1_fetch/src/map_utils.R")

county <- scmake("focus_geoms")

fil <- precip$time == precip$time[1]

p <- left_join(select(precip_spatial, id), 
               select(filter(precip, fil), -time), 
               by = "id")

plot(view_polygon, lwd = 3, main = precip$time[1])
plot(p["precip"], border = "transparent", add = TRUE)
plot(county$geometry, add = TRUE)
plot(bbox, lwd = 3, add = TRUE)
```
(slight mod of your code), I see
image

which looks good

1_fetch/src/fetch_precip_values.R
dblodgett-usgs: As we talked about on the phone -- this is a bit safer. And since we are joining it up to a set of spatial points, this seemed like a happy medium. I'm going to stick with this format and move on to process where I'll dump out a list of rasters with time stamp as the name of each raster.

1_fetch/src/fetch_precip_values.R
dblodgett-usgs: Didn't want to library(magrittr) I cleaned it up a big by pulling the origin out.

jread-usgs: You've already got it in there:
```yaml
packages:
  - sf
  - dplyr

```

dblodgett-usgs: So that means scipiper `library()`s them? Thought it was more like a package where they are available to import but not necessarily loaded...

aappling-usgs: Yeah, `remake` is sorta heavy-handed about packages - it does effectively `library()` anything you put in the `packages` list.

Issue #3, opened 2018-07-09
pipeline for gage information


aappling-usgs:
Site point information - gage coordinates, flood stage thresholds, and any additional site data by which we plan to order the sparklines - may come from multiple source datasets, so data pulls from these various datasets may need to be coordinated.

At least in recent vizzies, the availability of NWS flood stage information has largely determined the list of sites for which NWIS site information & stage data were pulled.

We also want some way to be able to specify which sites we want to include in the final movie, probably after some initial broader data pull that gives us the info we need to be able to choose that final site list.

lindsaycarr:
Common subsetting:
Exclude gages that don't have NWS flood stage.
Filter out gages not in viewbox.

Examples of custom filtering functions per storm:
Filter based on not being predicted to exceed flood stage.
Hand drawn polygon.
Filter out smaller gages (use drainage areas).
Hand picked (manual exclusion) - remove tidal gages, really messy looking ones, gages going offline

lindsaycarr:
After #42, there are 245 sites:

![image](https://user-images.githubusercontent.com/13220910/42598055-993728cc-8520-11e8-8e5b-a67d13139316.png)

aappling-usgs:
Nice, that's a lot of cleanup! might still be more gages than we'd want to show sparklines for, but gets us to a much more tractable number of possibilities

lindsaycarr:
I think this can be closed now. I don't think it's within scope of this sprint to work on pruning the gage list any more in the custom function.

PR #23, opened 2018-07-09
Basic gage info from NWIS

lindsaycarr:
Part of #3 (still need NWS flood info). I moved some of `1_fetch.yml` around because this step depends on the spatial steps, and the streamflow fetch depends on this.

For now, I just used `sample_n` to grab only 50 of the 3000+ sites. We need to have a greater discussion about how we are filtering sites down.


1_fetch/src/fetch_sites.R
```diff
@@ -0,0 +1,52 @@
...
+  state_cds <- unique(dataRetrieval::stateCdLookup(state_names))
+
+  # Cast wide net for all NWIS sites with stage data that fall within that bbox
+  sites_df <- data.frame()
+  for(cd in state_cds) {
+    sites_df_cd <- dataRetrieval::whatNWISdata(stateCd = cd, parameterCd = stream_params$stage, service = "uv")
+    sites_df <- rbind(sites_df, sites_df_cd)
+  }
+  sites_df <- dplyr::filter(sites_df, site_tp_cd == "ST")
+  sites_df <- dplyr::filter(sites_df, end_date >= as.Date(dates$dates$end))
```
jzwart: I had to do this too, but any idea why we have to do `dates$dates$...` when we've already subset to `dates` in the `viz_config`?

1_fetch/cfg/stream_params.yml
```diff
@@ -1,3 +1,4 @@
 flow:
   - '00060'
-  - '00072'
+stage:
+  - '00065'
```
jzwart: need to change code in `fetch_streamflow` if we want stage + discharge

1_fetch/src/fetch_sites.R
```diff
@@ -0,0 +1,52 @@
...
+}
+
+#' Subset the site list to ones relevant for the storm
+#'
+#' @param ind_file character file name where the output should be saved
+#' @param all_sites data.frame of sites pulled for the bbox
+subset_sites <- function(ind_file, all_sites) {
+
+  scipiper::sc_retrieve(all_sites)
+  all_sites <- readRDS(scipiper::as_data_file(all_sites))
```
jzwart: I think you can combine these lines to: `all_sties <- readRDS(sc_retrieve(all_sites))`

1_fetch/src/fetch_streamdata.R
```diff
@@ -0,0 +1,14 @@
+fetch_streamdata <- function(ind_file, sites, dates, stream_params, gd_config) {
+
+  scipiper::sc_retrieve(sites)
+  sites <- feather::read_feather(scipiper::as_data_file(sites))
+
+  message('  starting download of NWIS data at ', Sys.time())
+  data <- readNWISuv(
+    siteNumbers=sites$site_no, parameterCd=stream_params$stage,
```
jzwart: ah you got it, disregard my comment above

1_fetch/src/fetch_sites.R
lindsaycarr: I think @aappling-usgs fixed that here, but I haven't had a chance to merge those changes in just yet! https://github.com/USGS-VIZLAB/vizstorm-GIF/pull/24/commits/a58c5513b8f2039b5d5159d93fba36a36cbbca08

1_fetch/src/fetch_sites.R
lindsaycarr: 👍

PR #24, opened 2018-07-09
Timesteps placeholder

aappling-usgs:
placeholder function that creates a vector of timesteps for us to use in other functions.

later we'll want to tie the final choice of timesteps to the availability of data, especially precip data, but this should be enough to get us going for now.


2_process/src/choose_timesteps.R
```diff
@@ -0,0 +1,7 @@
+choose_timesteps <- function(ind_file, dates) {
+  # temporary quick version of creating timesteps - later we should also look to
+  timesteps <- seq(as.POSIXct(dates$start, tz='UTC'), as.POSIXct(dates$end, tz='UTC'), by=as.difftime(1, units='hours'))
```
dblodgett-usgs: Side note, precip will be in units of hours. -- do we want everything in UTC time to keep ourselves safe?

aappling-usgs: *units* of hours? like units=resolution, or quantity units, or hours since some initial time point?

without being entirely clear on the first part, i still feel pretty confident agreeing that everything in UTC is almost certainly the way to go =)

dblodgett-usgs:
Sorry... yeah, units as in resolution. NetCDF time units are "hours since ..." and they increment by 1.

PR #25, opened 2018-07-10
Process storm points

aappling-usgs:
* Switch storm ID to Hurricane Harvey
* Change assumed TZ from Puerto Rico (copied from Maria) to UTC (does that mean Maria was off? NHC metadata give the same definition of the one time field they bother to define for both Harvey and Maria)
* Extracted storm points from NHC data and projected to viz_config$projection
* Interpolated storm points relative to 2_process/out/timestamps.rds


jread-usgs:
This looks good, but I see these two coarse points that don't look like the interp spanned them correctly:

image

(green is coarse, red is interp). Looking closer at it too. 

Also, since you have updated the storm in this PR, would you like to update the view box too? Maybe `bbox: [-100, 22, -80, 34]`?

aappling-usgs:
Yes, will update the bbox for sure.

Taking a look at the interp. I haven't located that one point yet, but I think some variation should be expected in the number of interpolated (red) points that fall between each set of interpolated (green) points because the green points aren't even in time.

From within interpolate_storm_points.R, I ran this:
```
> interp_df <- tibble::as_tibble(unclass(coords_interp_sf))%>%mutate(type='interp')%>%select(DateTime, type)
> coarse_df <- tibble::as_tibble(unclass(coarse_points))%>%mutate(type='coarse')%>%select(DateTime, type)
> bind_rows(coarse_df, interp_df) %>% ggplot(aes(x=DateTime,y=type))+geom_point(alpha=.2) + xlim(range(interp_df$DateTime))
```
There are a few times, around Aug 26 and Aug 30, when the number of coarse/green points gets denser:
image

jread-usgs:
Oh, I didn't realize the green points weren't on even time steps. Makes sense now. I thought those were every 6 hours.

aappling-usgs:
Yeah, I hadn't thought about it much until you pointed it out, but so far I think the data are OK. Still much appreciate your careful inspection - this stuff can be hard to get right!

Also, how did you get your red and green dots to line up? I had to switch to ggplot because my naive base R approach of using `plot(..., add=TRUE)` seems to have updated the x and y lims from the first plot to the second one. Do I need to identify and fix the xlim and ylim, or is there a simpler way?

jzwart:
you could specify xlim and ylim in both plots or do `points()` for you second plot, which will keep first plot's x and y lims

aappling-usgs:
oh, right, `points()`! gollllyy am i rusty. thanks

jread-usgs:
Looks solved above, but I did:

```r
view_polygon = scipiper::scmake('view_polygon')
interp_pt <- readRDS('2_process/out/storm_points_interp.rds')
coarse_pt <- readRDS('2_process/out/storm_points_coarse.rds')
sp::plot(view_polygon)
sp::plot(interp_pt, col = 'red', add = TRUE)
sp::plot(coarse_pt, col = 'green', add = TRUE)
```

aappling-usgs:
Cool, thanks, that's really clean. And interesting that `sp::plot` will plot an `sf` object... Why did you choose that instead of `sf:::plot.sf` (i.e., just `plot()`)? I think I was able to get sf plotting functions to do the same thing with
```r
plot(sf::st_sf(view_polygon))
plot(coarse_pt, col='green', add=TRUE)
plot(interp_pt, col='red', add=TRUE)
```

jread-usgs:
yeah, I was using `sp::plot` because I restarted R and didn't have anything loaded and that worked for me when `plot()` wasn't working (it was before I closed my session)...

PR #31, opened 2018-07-10
Precip as a raster!!

dblodgett-usgs:
To work with the output do something like:

```r
rasters <- readRDS("2_process/out/precip_rasters.rds")
view_polygon <- scmake("view_polygon")

county <- scmake("focus_geoms")

library(RColorBrewer)

cuts=c(5,10,20,40,80) #set breaks
pal <- colorRampPalette(c("grey","green", "purple", "pink"))

plot(view_polygon, lwd = 3, main = names(rasters)[70])
plot(county["geometry"], borders = "grey", add = T)
plot(rasters[[70]], breaks = cuts, col = pal(5), add = TRUE)
```

Output of that looks like:
screen shot 2018-07-10 at 12 10 25 pm

And just to double check that things look to line up right:
screen shot 2018-07-10 at 12 13 03 pm

Each raster in the list this outputs looks like:
```
> rasters[1]
$`2017-08-25 01:00:00`
class       : RasterLayer 
dimensions  : 569, 1024, 582656  (nrow, ncol, ncell)
resolution  : 3003.295, 3002.708  (x, y)
extent      : 604126.2, 3679500, -2441803, -733262  (xmin, xmax, ymin, ymax)
coord. ref. : +proj=laea +lat_0=45 +lon_0=-100 +x_0=0 +y_0=0 +a=6370997 +b=6370997 +units=m +no_defs 
data source : in memory
names       : layer 
values      : 0, 54.5  (min, max)
```


dblodgett-usgs:
Wheee!!
screen shot 2018-07-10 at 1 12 19 pm

```r
rasters <- readRDS("2_process/out/precip_rasters.rds")
view_polygon <- scmake("view_polygon")

county <- scmake("focus_geoms")

library(RColorBrewer)

cuts=c(10, 100, 200, 300, 400, 800, 1500) #set breaks
pal <- colorRampPalette(c("grey", "green", "purple", 
                          "pink", "black", "red", "yellow"))

plot(view_polygon, lwd = 3, main = names(rasters)[length(rasters)])
plot(county["geometry"], borders = "grey", add = T)
plot(rasters[[length(rasters)]], breaks = cuts, 
     col = pal(6), alpha = 0.7, add = TRUE)
```

2_process/src/process_precip.R
```diff
@@ -0,0 +1,68 @@
...
+    ungroup()
+
+  ncol <- view_config$width
+
+  bbox <- sf::st_bbox(view_polygon)
+
+  y_range <- as.numeric(bbox$ymax - bbox$ymin)
+  x_range <- as.numeric(bbox$xmax - bbox$xmin)
+
+  aspect_ratio <- x_range / y_range
```
aappling-usgs: aspect_ratio is present in viz_config.yml; is there any reason it'd be different from `view_config$aspect` by the time we get here? though as noted above, i'm hoping we can switch to `view_config$width` and `view_config$height` instead of keeping aspect in the config, in which case computing it from width and height would be fine. But still, can we skip the y_range and x_range lines, or are those needed for some reason?

viz_config.yml
```diff
@@ -5,6 +5,7 @@ dates:
 bbox: [-100, 22, -80, 34]
 projection: "+proj=laea +lat_0=45 +lon_0=-100 +x_0=0 +y_0=0 +a=6370997 +b=6370997 +units=m +no_defs"
 aspect: 1.8
```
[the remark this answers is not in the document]
aappling-usgs: lindsay recommends `height: 512` to match `width: 1024`

2_process/src/process_precip.R
dblodgett-usgs: Yeah. This is messy from how I had it before. I'll get height / width in to view_config and clean house a bit.

aappling-usgs: :+1:

Issue #13, opened 2018-07-09
coordinating storm animation frame function


aappling-usgs:
generates a single animation frame (GIF format), with timestep determined by a function argument, showing the state of the storm and precip and stream stages

this function shouldn't do much (or any) actual plotting itself. instead, it should call other little viz functions - basemap, legend + position of legend, sparklines panel + position of sparklines, etc. It _is_ the job of this function to determine where each of those things should be positioned, based mainly on info in viz_config.yml.

aappling-usgs:
i think we're on a roll with this one now. jordan persuaded me that this function should do even less than i originally proposed, so now most of the work rests with the prep functions that get passed to this function.

Issue #27, opened 2018-07-10
task table(s) for animation frames


aappling-usgs:
Let's use a scipiper task table or two to save image files for the animation frames generated by the functions described in #13 and #26. I think it may be simplest to have one task table for #13 and one for #26...but also be thinking about how these frames are going to get bundled up again into a single movie after they've been generated.

Rows=tasks of the task table should be timesteps. There could be one column=step, which would be to generate an animation frame, or there could be multiple columns=steps where each column is to generate an animation frame for the same timestep but with different resolution, aspect ratio, image format, etc -
 whatever is needed to allow us to produce all the different types of GIFs/movies we want to produce.

jzwart:
Should this live in `2_process` or `6_visualize`? or make a new folder like `3_gif_tasks` or something?

aappling-usgs:
`6_visualize`

aappling-usgs:
We're rolling with the task table now.

PR #33, opened 2018-07-10
prototype create_storm_frame

aappling-usgs:
A preliminary effort toward #13. This won't create anything at the moment but I hope will give a sense of how I think this plot coordination function can work once we plug in the pieces. The job of `create_storm_frame` is to decide where each component lives and then tell all the components to call themselves. (component = basemap, legend, etc.)


6_visualize.yml
```diff
@@ -10,8 +10,22 @@ file_extensions:
...
   6_visualize:
     depends:
+
+  # create_storm_frame will soon be called for all frames from within
+  # a task table. just calling it once here to illustrate
+  storm_frame_config:
+    command: viz_config[I(c('bbox','projection','width','height','focus_geoms','secondary_geoms'))]
+  6_visualize/out/example_storm_frame.png:
+    command: create_storm_frame(
+      png_file=target_name, timestep=I('2017-09-26 11:00:00'), config=storm_frame_config,
```
jzwart: I was just thinking about this for the task tables #27 , but how many frames do we want for the gif? I guess this could be stored in `viz_config` as `frame_time_interval`. Any idea what the interval should be for this initial gif? 6 hours?

aappling-usgs: At least for now, how about we have as many frames as there are timesteps in `2_process/out/timesteps.rds`?

jzwart: sounds good

jread-usgs: I was thinking hourly (frame per hour) as a default, but we'd want the ability to do every 2 hrs, 3 hrs, 6hrs etc as needed

aappling-usgs: Agreed, but I think we could do that subsetting earlier, in `choose_timesteps()`, and then have lighter-weight data to throw around during the visualization phase. Would be a pretty simple addition to viz_config.yml (something like `time_resolution_hrs: 3`) and modification to `choose_timesteps()` (just take every 3rd of the hourly precip timestamps, or get a little fancier with the POSIXct if needed)

PR #36, opened 2018-07-10
Custom gage subset

lindsaycarr:
Create the "custom gage filtering" function which would change from viz to viz and add to process step. Now, dependencies are circular between process and filter, so you need to run `scmake(c('1_fetch', '2_process'))` to build the whole thing.

More of #3. Using National Weather Service flood stage as a filter is coming in a separate PR.


1_fetch/src/fetch_sites.R
```diff
@@ -15,37 +15,21 @@ fetch_sites_from_sf <- function(ind_file, sf_object, dates, stream_params) {
...
-  # subset the sites from the wide net cast to ones relevant to the storm
-  # subset criteria TBD
-  sites_info <- dataRetrieval::readNWISsite(siteNumbers = all_sites)
-
-  #############********* temporarily grab only a few sites *********#############
-  set.seed(10)
-  sites_info <- dplyr::sample_n(sites_info, 50)
-  #############*********
+  # return only the site numbers
+  sites <- sites_df$site_no
```
dblodgett-usgs: Should we maybe convert this to an sf data.frame? 
`sf::st_as_sf(sites, coords = c("dec_long_va", "dec_lat_va"), crs = 4326)`

2_process/src/filter_sites_custom.R
```diff
@@ -0,0 +1,23 @@
...
+
+  sites <- readRDS(scipiper::sc_retrieve(sites_file))
+
+  # subset the sites from the wide net cast to ones relevant to the storm
+  # subset criteria TBD
+  sites_info <- dataRetrieval::readNWISsite(siteNumbers = sites)
+
+  sites_info_subset <- sites_info %>%
+    # only include sites with large drainage areas
+    filter(drain_area_va > quantile(drain_area_va, 0.75, na.rm=TRUE))
```
dblodgett-usgs: Have you played with the 0.75 input here? Should we parameterize that?

2_process/src/filter_sites_custom.R
```diff
@@ -0,0 +1,23 @@
...
+
+  # write the data file and the indicator file
+  data_file <- as_data_file(ind_file)
+  feather::write_feather(sites_info_subset, data_file)
```
dblodgett-usgs: Why a mix of feather and rds?

1_fetch/src/fetch_sites.R
lindsaycarr: Hmm I think it would be easier to do this after the sites are confirmed. So maybe as a processing step right after the custom sites function?

2_process/src/filter_sites_custom.R
aappling-usgs: i don't have a preference for this project and am OK with having both

PR #38, opened 2018-07-11
An initial pattern for building a single frame for the GIF

jread-usgs:
storm_frame
Generates this for starters


5_vizprep/out/storm_point_[20170826-11]_fun.rds.ind
```diff
@@ -0,0 +1,2 @@
+indication_time: 2018-07-11 13:13:12 UTC
```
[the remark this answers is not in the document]
Comment (aappling-usgs) [reply 5_vizprep/out/storm_point_[20170826-11]_fun.rds.ind]: but since david just resolved https://github.com/USGS-R/scipiper/issues/54, i'll leave this in here so we can delete the other ones and rebuild this on in the future. expecting a change from `indication_time` to `hash` in that future version of this file; i assume indication_time was from some earlier version of the function that created this file.

6_visualize.yml
```diff
@@ -27,19 +27,18 @@ targets:
   # into an RDS file with a corresponding .rds.ind indicator file.
 
   # right here we will create and run a task table that will be responsible for generating
-  # the components, e.g., '6_visualize/out/style/storm_point_20170926_11.rds.ind' through
-  # '6_visualize/out/style/streamdata_20170926_11.rds.ind', one per column=step. and then the final
+  # the components, e.g., '5_vizprep/out/storm_point_20170926_11.rds.ind' through
+  # '5_vizprep/out/streamdata_20170926_11.rds.ind', one per column=step. and then the final
```
aappling-usgs: i think you're proposing that the task table generated in 6_visualize puts some files into 5_vizprep. that seems like an unnecessary bending of our usual project structure practices; i'd rather keep these output files in 6_visualize if you can be persuaded

6_visualize.yml
```diff
@@ -27,19 +27,18 @@ targets:
...
-      '6_visualize/out/style/focus_geoms.rds.ind',
-      '6_visualize/out/style/secondary_geoms.rds.ind',
-      '6_visualize/out/style/storm_line.rds.ind',
-      '6_visualize/out/style/storm_point_20170926_11.rds.ind',
-      '6_visualize/out/style/precip_raster_20170926_11.rds.ind',
-      '6_visualize/out/style/streamdata_20170926_11.rds.ind')
+      '5_vizprep/out/view_fun.rds.ind',
+      '5_vizprep/out/basemap_fun.rds.ind',
+      '5_vizprep/out/storm_line_fun.rds.ind',
+      '5_vizprep/out/storm_point_-20170826_11-_fun.rds.ind')
```
aappling-usgs: no good just using a dash to separate the days from the hours, huh? e.g., `5_vizprep/out/storm_point_20170826-11_fun.rds.ind`? that's how i would have done this, to preserve the ability to really easily parse the string with a strsplit on `_`. what's your standard code for parsing with `[` and `]` in some pieces of the filename?

5_vizprep.yml
```diff
@@ -0,0 +1,42 @@
...
+file_extensions:
+  - feather
+  - ind
+
+sources:
+  - 5_vizprep/src/prep_plot_funs.R
+
+targets:
+
+  5_vizprep:
```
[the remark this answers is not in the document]
jzwart: That's what I was imagining. Task table will generate dynamic targets that will depend on non dynamic targets in `5_vizprep.yml`

5_vizprep/out/storm_point_[20170826-11]_fun.rds.ind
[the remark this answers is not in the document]
Comment (jread-usgs) [reply 5_vizprep/out/storm_point_[20170826-11]_fun.rds.ind]: I'd like to go back to the `[time]` notation because I think it makes it more explicit that we are parsing the content w/in those brackets and easier to read. I used `-` as a placeholder with the hope that I could change it back when that vizlab update was made.

6_visualize.yml
jread-usgs: I want it to use `[20170826-11]` and switched that to `-` and `_` as a placeholder

jread-usgs: to get what is within the first set of brackets:
```r
gsub("\\-|\\-", "", regmatches(filename, gregexpr("\\-.*?\\-", filename))[[1]][1])
```

6_visualize.yml
jread-usgs: I did this before looking at the task table PR, so I was envisioning a visualization prep stage that would precede this visualize stage. If we are going a different route, I think we can change this easily.

5_vizprep.yml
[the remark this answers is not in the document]
jread-usgs: I was thinking there would be a dynamic task table in `5_` for these time-specific functions. Bad idea?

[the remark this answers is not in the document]
jzwart: oh I see to create all the `storm_point_[date].rds`, etc.. in `5_vizprep/out`. I think that's probably good actually

6_visualize.yml
aappling-usgs: thanks

6_visualize.yml
aappling-usgs: OK, let's change it. You proposed a viz prep stage yesterday, so that does not surprise me, but then i had suggested on the phone (and want to keep pushing for this) that all of the dynamic files should live in a single task table, whether they're prep or actually create a png, because of the funkiness about having task-specific targets in one task file depend on task-specific targets in a different task file. So that's why my comment above (deleted lines 30 and 31) had them in a subfolder within phase 6.

5_vizprep/out/storm_point_[20170826-11]_fun.rds.ind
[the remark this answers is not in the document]
Comment (aappling-usgs) [reply 5_vizprep/out/storm_point_[20170826-11]_fun.rds.ind]: Yes. go for it.

5_vizprep.yml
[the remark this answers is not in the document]
aappling-usgs: I advocated yesterday and am going to continue advocating for there to be just one task table, all in phase 6, to avoid funky dependencies across task tables.

6_visualize.yml
jread-usgs: Yep, agreed. can change. I do remember that, but hadn't thought about how to make the connection w/ the other task table yet. Doing that now that I see the other work and your suggestions here.

aappling-usgs: :+1:

PR #39, opened 2018-07-11
View polygon ind

aappling-usgs:
i think this will be everything we need to deal with the OS differences in `sf::st_transform`

### DIFF
--- a/1_fetch.yml
+++ b/1_fetch.yml
@@ -29,24 +29,18 @@ targets:
 
   1_fetch:
     depends:
-      - 1_fetch/out/streamdata.rds.ind
-      - 1_fetch/out/storm_track.zip.ind
       - view_polygon
+      - 1_fetch/out/storm_track.zip.ind
       - 1_fetch/out/all_sites.rds.ind
+      - 1_fetch/out/streamdata.rds.ind
+      - 1_fetch/out/precip_spatial.rds.ind
+      - 1_fetch/out/precip_values.rds.ind
 
   # -- config --
   dates:
     command: viz_config[[I('dates')]]
   stream_params:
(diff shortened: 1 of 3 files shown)


aappling-usgs:
I'm intentionally not commiting new or changed build/status files because i'd like @dblodgett-usgs or @jread-usgs to rebuild the spatial stuff (mainly for the sake of the precip data) on one of their Macs

1_fetch.yml
```diff
@@ -29,51 +29,58 @@ targets:
...
 
   view_config:
     command: viz_config[I(c('bbox', 'projection', 'width', 'height'))]
 
+  1_fetch/out/view_polygon.rds.ind:
+    command: post_view_polygon(target_name, view_config)
+  1_fetch/out/view_polygon.rds:
+    command: gd_get('1_fetch/out/view_polygon.rds.ind')
+  view_polygon:
+    command: get_view_polygon('1_fetch/out/view_polygon.rds.ind')
```
lindsaycarr: I know that you needed to make secondary_geoms and focus_geoms have files and indicator files in order to get around the OS issues with rebuilding, but does that mean we can't do the third step here for those and get them back as an object to pass around?

aappling-usgs: i think we could. but is it worth it for secondary_geoms and focus_geoms? those get used a lot less than view_polygon. that's why i didn't do it for all three.

aappling-usgs: taking it to the extreme, every file in this whole project could have that third step. but then everybody would always need to download every file, reducing some of the benefit of the shared cache. so i think drawing the line closer to very few cases like `view_polygon` is a good idea.

PR #40, opened 2018-07-11
rebuild

dblodgett-usgs:


aappling-usgs:
hmm, i was expecting to see the view_polygon and geoms in here, too. i think i should have sugggested several additional scmake calls:
```r
scmake('1_fetch/out/view_polygon.rds.ind', force=TRUE)
scmake('1_fetch/out/secondary_geoms.rds.ind', force=TRUE)
scmake('1_fetch/out/focus_geoms.rds.ind', force=TRUE)
```
could you run those, too?

aappling-usgs:
@jzwart i haven't tested recently so am not 100% sure, but i'm pretty confident that it would _not_ force a rebuild of those dependencies, because `force` applies to the target and not its dependencies. I see how that could be useful for these group targets, though, and we could consider making a special case for them.

PR #42, opened 2018-07-11
Add nws data, filter by missing stage, & filter by focus geoms + view polygon

lindsaycarr:
This has some back-and-forth dependencies between `2_process` and `1_fetch`. The rest of #3.

`1_fetch/out/nws_stage_sites.rds.ind` step takes a long time (especially when the view is large - takes about 30 min at current view size).

Subsetting by view looks like this (this image doesn't appear anywhere in this code, only used to illustrate what's going on in this PR): 
1. Blue are all of the sites that come back from querying NWIS by states available in focus_geom and is the state of the site list at `1_fetch/out/all_sites.rds`.
2. Red are the sites subsetted from `1_fetch/out/all_sites.rds` that fit within the box defined by `view_polygon`.

Red sites are then used to get National Weather Service data and filter on the existence of flood stage (see `1_fetch/out/nws_stage_sites.rds.ind`). Then, those sites are passed through the custom filtering function, see `2_process/out/storm_sites_custom.rds.ind`. The last step is to use the sites that are returned from the custom filtering to subset the site data that has NWS flood stage and location information, see `2_process/out/storm_sites.rds.ind` (this is kept separate so that all future custom functions need to worry about is returning a data.frame with a site_no column of the appropriate site numbers).

image

I built this locally using only 10 sites to get National Weather Service data to make sure it ran. I am currently building with 1,967 because that is what is returned from `2_process/out/storm_sites_geom.rds.ind`. I will push the ind and build/status files up when that finishes, but I want to get someone looking at this.


1_fetch/src/fetch_sites.R
```diff
@@ -29,12 +36,82 @@ fetch_sites_from_sf <- function(ind_file, sf_ind, dates, stream_params) {
...
+  i_count <- 0
+  for(i in site_nums_to_query){
+
+    if(is.na(i)) {
+      i_count <- i_count + 1
+      print(paste(i_count, "/", length(site_nums_to_query)))
+      next
+    }
+
+    url_site <- paste0("https://water.weather.gov/ahps2/hydrograph_to_xml.php?gage=",i,"&output=xml")
```
aappling-usgs: oof, i see why this loop takes so long! i think that's ok, though, because we should really only have to do this once per storm (unless we need to update our viewbox repeatedly...hmm...). and also, i don't know of a faster way to do it, so there we are.

2_process/src/filter_sites_custom.R
```diff
@@ -2,22 +2,23 @@
...
 
   # subset the sites from the wide net cast to ones relevant to the storm
-  # subset criteria TBD
-  sites_info <- dataRetrieval::readNWISsite(siteNumbers = sites)
+  # subset criteria TBD for each storm
+  sites_info <- dataRetrieval::readNWISsite(siteNumbers = sites_df$site_no)
 
   sites_info_subset <- sites_info %>%
     # only include sites with large drainage areas
     filter(drain_area_va > quantile(drain_area_va, 0.75, na.rm=TRUE))
```
aappling-usgs: this is fine for this demo custom function, but we should keep a close eye on `drain_area_va` when we get to real storms, because i don't think NWIS reliably supplies drain_area_va for every site.

aappling-usgs:
And we're waiting on a commit of  2_process/out/storm_sites_geom.rds.ind and its build/status file, right?

1_fetch/src/fetch_sites.R
jread-usgs: I think there is a waterwatch service or web table we could potentially scrape for this...?

lindsaycarr:
True!

Should I almost commit the updated `1_fetch/out/streamdata.rds.ind` and corresponding build/status file? That was updated because the sites being fed into it updated.

aappling-usgs:
Yes, go ahead and commit the streamdata.rds.ind + build/status file, too. Because (1) you've gone to the work, (2) i don't think anybody else is messing with that bit right now, and (3) you've already posted it to Drive, sounds like, which we should use as our primary guideline for when to commit ind/buildstatus files

1_fetch/src/fetch_sites.R
aappling-usgs: Happy to see this get addressed in a future PR if such service/table does exist

aappling-usgs:
Yeah, I think we learned more about that today :grimacing: . Nice to be exercising scipiper in this viz/collaborative context - I think it'll make us stronger and faster for future projects!

Issue #11, opened 2018-07-09
pipeline for river network


aappling-usgs:
non-MVP but woudl be cool

dblodgett-usgs:
Here's a concept to consider. I just used the NLDI to get the upstream mainstem and downstream mainstem for each gage. For the sake of seeing what it looks like, I just plotted the upstream in red and downstream in blue.
![screen shot 2018-07-10 at 3 16 22 pm](https://user-images.githubusercontent.com/1492803/42535160-42cd079e-8454-11e8-8f22-e98810028daa.png)

I think this could be a good and simple way to demonstrate how the gages relate to the drainage network. It doesn't attempt to represent all the hydrography, but does show how the gages relate to a broader hydrologic network. 

Thoughts?

aappling-usgs:
I sorta love it - it's definitely the right density and shows how the gages are or are not connected to one another and to the landscape.

In case this wasn't already your intent, I think we should use just one color for the final hydrography lines, or at least not make an effort to distinguish visually between upstream and downstream. It's great for our discussion but I think would distract an uninitiated audience.

At the same time that I'm very happy with how this looks, I'm worried that somebody is going to think these are the major rivers (or our failed attempt to capture the major rivers). The disadvantage/challenge of a GIF-only viz is we don't get any text to support our communication, so it really really needs to stand alone.

As a next iteration, maybe we could try showing both:
1. in dark blue: the largest rivers in the network, regardless of whether they have our gages on them, and
2. in lighter blue: all rivers with gages on them, as in your prototype above but not re-plotting those that are both large and have gages

@dblodgett-usgs and others, what do you think - do you agree with my concern? do you think that my proposal above would get us out of trouble, or would it still be confusing?

jread-usgs:
I agree that adding the up/downstream style difference adds a new dimension to the viz that may be confusing for the target audience. I like the proposal Alison has as a next step. Nice work Dave!

dblodgett-usgs:
Yes, my intent with the color was for this issue only. I was imagining the same styling throughout for the viz.

I like your proposal, will work that up once I figure out a good way to seed "largest rivers in the network". I have an idea.

jzwart:
This is awesome

dblodgett-usgs:
This took a lot of doing, but I got it. 

[...]

PR #45, opened 2018-07-11
River network geometry

dblodgett-usgs:
As noted in #11 - I want to rework how to major rivers are selected. But this is ready for an initial merge.


dblodgett-usgs:
Note that I have not tried to build what is contributed in 6_prep.

1_fetch/src/fetch_river_geometry.R
```diff
@@ -0,0 +1,74 @@
...
+
+
+sf_converter <- function(x)
+  sapply(x, function(x) try(sf::read_sf(x), silent = TRUE),
+         USE.NAMES = TRUE, simplify = FALSE)
+
+navigate_nldi <- function(f_id, f_source, mode = "UM",
+                          d_source = NULL, distance = NULL) {
+
+  url <- paste("https://cida.usgs.gov/nldi", f_source, f_id, "navigate", mode, d_source,
```
jread-usgs: works the same, but I find `sprintf` to be a lot more readable for these kinds of things:
```r
url <- sprintf"https://cida.usgs.gov/nldi/%s/%s/navigate/%s/%s", f_source, f_id, mode, d_source)
```
not worth a change, but just a thought.

6_vizprep/src/prep_rivers_fun.R
```diff
@@ -0,0 +1,12 @@
+
+prep_rivers_fun <- function(ind_file, river_ind){
+
+  sf_rivers <- readRDS(sc_retrieve(river_ind))
+
+  plot_fun <- function(){
+    plot(sf_rivers$sf_gage_rivers$geometry, col = "deepskyblue", lwd = .5, add = TRUE)
```
jread-usgs: 👍

1_fetch/src/fetch_river_geometry.R
dblodgett-usgs: yeah... I always think to myself, I should really be templating this, but never do.

PR #47, opened 2018-07-12
Working version of multiple frame creation

jread-usgs:
This generates frames that can be combined into a gif:
test_animation

I think I would like to discuss this, as I think it is going to quickly become unwieldy when we expand timesteps. 

In designing how we'd partition plotting and plotting-prep, I thought it would be good to target quick iteration on single frames, because getting the style and layout right seems to be a lengthy process w/ lots of small iterations. I didn't want a large processing cost to be paid each time a style change was made (we can focus on single frames while iterating w/ `scmake('6_visualize/tmp/gif_frame_a_20170825_03.png')` for example). 

Given that, I thought it would be beneficial to slice up the time frames and prep them for plotting, but I wasn't thinking about the impact of round-tripping to gdrive for these little files. Yes the indicator files are smaller than the plot function .rds, but the .rds aren't that big (~40K). 

Anyhow, I don't want to get too wrapped up in finding an ideal solution here, as this is workable. But looking for input.


aappling-usgs:
> Given that, I thought it would be beneficial to slice up the time frames and prep them for plotting, but I wasn't thinking about the impact of round-tripping to gdrive for these little files. Yes the indicator files are smaller than the plot function .rds, but the .rds aren't that big (~40K).

Agreed - we don't want to post to google drive for these little files! Let's have these be local files, so anybody who wants them needs to build them him/herself. Use `sc_indicate(ind_file, data_file=data_file)` instead, or - even better, i think, because these are small and quickly hashable files - skip the generation of indicator files entirely. I'll make a suggestion for how to do it in the files in a minute.

aappling-usgs:
Remember to assign/propose a reviewer, please

6_visualize.yml
```diff
@@ -48,13 +56,40 @@ targets:
       tmp=I('6_visualize/tmp'),
       log=I('6_visualize/log'))
 
+  6_vizprep/out/view_fun.rds.ind:
```
aappling-usgs: I'd prefer it if the .yml filename and the directory of these outputs matched. Renaming the .yml to 6_vizprep.yml would be one option, and I think the still-better option would be, given that we're generating both preparatory functions and final png files within this .yml, that we put all the outputs into 6_visualize.

You and I probably agree that it's too many files to be dumping into a single `out` folder all together, so I propose that we create subfolders within `6_visualize/out`. To get super organized, we could have:
* `6_visualize/out/prep_static` (for view_fun.rds, basemap_fun.rds, storm_line_fun.rds, rivers_fun.rds, and watermark_fun.rds, and their corresponding .ind files)
* `6_visualize/out/prep_dynamic` (for all xx_fun.rds that are timestep specific)
* `6_visualize/out/frames` (for all .png output files)
* `6_visualize/out/gif` (for the final thing)

Or we could go with something a little simpler if you prefer - I don't care much about how we divide the subfolders, really just want that top folder to be aligned with the yml. A simpler example:
* `6_visualize/out/prep`
* `6_visualize/out/final`

6_visualize/src/create_gif_tasks.R
```diff
@@ -2,17 +2,18 @@
...
-create_gif_tasks <- function(timestep_ind, folders, view_ind, basemap_ind, storm_line_ind, storm_cfg){
+create_gif_tasks <- function(timestep_ind, folders, view_ind, basemap_ind, storm_line_ind, river_lines_ind, watermark_ind, storm_cfg){
 
-  timestep <- readRDS(sc_retrieve(timestep_ind))
+  message('subsetting times to simplify for now')
+  timestep <- readRDS(sc_retrieve(timestep_ind))[1:10]
 
-  cfgs <- c('a','b','c') # dummy placement for different configurations; will eventually be configurations that hold information about size, aspect, ect...
+  cfgs <- c('a') # for now, just use one config, since > 1 results in duplication of input files
+  #,'b') # dummy placement for different configurations; will eventually be configurations that hold information about size, aspect, ect...
```
aappling-usgs: :+1:

6_visualize.yml
```diff
@@ -48,13 +56,40 @@ targets:
...
+
+  6_vizprep/out/rivers_fun.rds.ind:
+    command: prep_rivers_fun(target_name, '1_fetch/out/river_geometry.rds.ind')
+  6_vizprep/out/rivers_fun.rds:
+    command: gd_get('6_vizprep/out/rivers_fun.rds.ind')
+
+  6_vizprep/out/watermark_fun.rds.ind:
+    command: prep_watermark_fun(target_name, '6_visualize/in/usgs_logo_black.png')
+  6_vizprep/out/watermark_fun.rds:
+    command: gd_get('6_vizprep/out/watermark_fun.rds.ind')
```
aappling-usgs: Depending on how long these function.rds files take to generate, and how frequently we need to do it, I think we may want to leave these 5 files off of Google Drive (in addition to taking the .png frames off of Drive). We'd be making the decision to not share a cache for these products; instead, the shared cache would take everyone through the fetch and process steps together but anybody who wanted one of these files would need to build it locally. The benefit would be speed of iteration in this 6th phase. Cutting off the shared cache means it's trickier to add a shared-cache file later in the pipeline, i.e., sharing the final .gif as a shared-cache file or sharing other output files that depend on these files...but because we don't have any phases after this one, that truncation of the shared cache shouldn't be a problem for us.

6_visualize/src/create_gif_tasks.R
```diff
@@ -24,6 +25,29 @@ create_gif_tasks <- function(timestep_ind, folders, view_ind, basemap_ind, storm
     paste(strs, collapse=sep)
   }
 
+  point_frame <- scipiper::create_task_step(
+    step_name = 'point_frame',
+    target_name = function(task_name, step_name, ...){
+      cur_task <- dplyr::filter(rename(tasks, tn=task_name), tn==task_name)
+      sprintf('6_vizprep/out/storm_point_[%s].rds.ind', cur_task$date_hour)
```
aappling-usgs: For these point frames, let's cut straight to the data file (`.rds` only) rather than adding the shared-cache layer (`.rds.ind`). 

The benefit of just writing a data file, and not posting to google drive, is that we save a little bit of file writing time and a lot of file posting/getting time.

The drawback is that we will not be sharing these .rds files across the team. Instead, what this means is that anybody who wants to see these .rds files, or anything that depends on them, will need to build them themselves. This constraint will also apply to files that depend on these .rds files - in particular, the .png files will need to follow a similar pattern (.png only, no .png.ind files). But I think that's OK and what we want for this project.

6_visualize/src/create_gif_tasks.R
```diff
@@ -24,6 +25,29 @@ create_gif_tasks <- function(timestep_ind, folders, view_ind, basemap_ind, storm
...
+    step_name = 'get_point_frame',
+    target_name = function(task_name, step_name, ...){
+      cur_task <- dplyr::filter(rename(tasks, tn=task_name), tn==task_name)
+      sprintf('6_vizprep/out/storm_point_[%s].rds', cur_task$date_hour)
+    },
+    command = function(task_name, ...){
+      cur_task <- dplyr::filter(rename(tasks, tn=task_name), tn==task_name)
+      sprintf("gd_get('6_vizprep/out/storm_point_[%s].rds.ind')", cur_task$date_hour)
+    }
+  )
```
aappling-usgs: So we can skip this whole `get_point_frame` step entirely.

6_visualize/src/create_gif_tasks.R
```diff
@@ -32,23 +56,27 @@ create_gif_tasks <- function(timestep_ind, folders, view_ind, basemap_ind, storm
     },
     command = function(task_name, ...){
       cur_task <- dplyr::filter(rename(tasks, tn=task_name), tn==task_name)
+
       psprintf(
         "create_storm_frame(",
-        "png_file=target_name",
+        "png_file=target_name,",
```
aappling-usgs: I guess we should rename this step from `gif_frame` to `png_frame` or `animation_frame` to be accurate.

6_visualize/src/create_gif_tasks.R
```diff
@@ -32,23 +56,27 @@ create_gif_tasks <- function(timestep_ind, folders, view_ind, basemap_ind, storm
...
   )
 
   gif_task_plan <- scipiper::create_task_plan(
     task_names=tasks$task_name,
-    task_steps=list(gif_frame),
+    task_steps=list(point_frame, get_point_frame, gif_frame),
     add_complete=FALSE,
+    final_steps='gif_frame',
     ind_dir=folders$log)
 }
```
aappling-usgs: Below - I think we should avoid git committing the frame .png files.

6_vizprep/src/prep_plot_funs.R
```diff
@@ -13,3 +13,28 @@ write_put_fun <- function(plot_fun, ind_file){
   saveRDS(plot_fun, data_file)
   gd_put(ind_file, data_file)
 }
```
aappling-usgs: If we go with the above suggestions, we can eliminate the need for `write_put_fun` entirely because we'll just be doing a `saveRDS(plot_fun, data_file)` without a `gd_put` or any need for an `ind_file` variable.

remake.yml
```diff
@@ -1,7 +1,7 @@
 target_default: 6_visualize
 
 include:
-  - 6_visualize.yml
+  - 6_gif_tasks.yml
```
aappling-usgs: Once we cut the tasks out of the shared cache system, I think we're going to want to keep 6_gif_tasks.yml disconnected from the main remake.yml. I can help write a `loop_tasks()` function that will save us from most of the extra typing (`remake_file='6_gif_tasks.yml'` in calls to `scmake`) that might otherwise result from this change.

6_visualize.yml
jread-usgs: I agree, and I was planning to move things around to avoid that folder. But I also considered my approach to be pretty flawed (hence this PR and questions in it), so I didn't want to re-organize w/o input on the overall approach.

6_visualize.yml
jread-usgs: I see. Good idea. Can you show me what that looks like to implement?

6_visualize/src/create_gif_tasks.R
jread-usgs: Sounds good. I think that is as simple as just cutting out that `.ind` step here?

6_visualize/src/create_gif_tasks.R
jread-usgs: will do

6_visualize.yml
aappling-usgs: OK great. I was hoping this wasn't going to be an extended discussion =). But know that I'd be willing to have it if you didn't agree.

6_visualize/src/create_gif_tasks.R
aappling-usgs: yep, pretty much, and skipping the call to gd_put

6_visualize.yml
jread-usgs: Yes, understood, but I am wondering instead how this works w/ connecting the different remake files. I should've been clear about that. I am in the process of setting these up as .rds and will leave the .yaml connections alone for the time being.

jread-usgs: Now that I think of it, if we are skipping the cache, is there a reason to use RDS instead of having this function return `plot_fun` and skip the save/readRDS? I am still getting used to when not to use objects vs rds.

aappling-usgs: Hmm, good point about RDS vs objects! I am also still getting used to that. I think you're right that an object would be fine...let's try it as objects, and we can always switch back if needed.

jread-usgs: Yeah, I think the reasons I want to do that: 1) simplicity, 2) it makes it really clear that it is outside of the scope of gdrive/cache.

aappling-usgs: About connecting the different remake files...I might be misinterpreting, but I think what we want is to (1) cut off the shared cache by switching away from .ind files and toward .rds files or R objects, and (2) keep going with the remake pipeline through phase 6 to the end in the same way as we would for a non-shared-cache pipeline. Therefore:

* This file, 6_visualize.yml, gets the same sort of group target as the 1_fetch.yml and 2_process.yml, but the dependencies are files or objects. Like this:
```yml
target_default: 6_visualize

includes:
  - 2_process.yml

targets:
  6_visualize:
    depends: 
      - basemap_fun
      - watermark_fun
      - [etc]
```
and then the main remake file, `remake.yml`, depends on `6_visualize`:
```yml
target_default: 6_visualize

include:
  - 6_visualize.yml

targets:
[empty]
```
The funny bit comes in at the task plan combine phase, which we'll deal with shortly.

Is that more the type of answer you were looking for?

jread-usgs: Yes, got it. Makes sense.

6_visualize.yml
```diff
@@ -48,13 +45,25 @@ targets:
...
+
+  storm_line_fun:
+    command: prep_storm_line_fun("2_process/out/storm_line.rds.ind")
+
+  rivers_fun:
+    command: prep_rivers_fun('1_fetch/out/river_geometry.rds.ind')
+
+  watermark_fun:
+    command: prep_watermark_fun('6_visualize/in/usgs_logo_black.png')
+
```
aappling-usgs: nice, very clean!

6_visualize/log/6_gif_tasks.ind
```diff
@@ -0,0 +1,192 @@
...
+6_visualize/tmp/gif_frame_a_20170831_02.png: 3b022323a4d1a4e7b88b4aab45791e27
+6_visualize/tmp/gif_frame_a_20170831_03.png: f3d14d0119a667386a9dfb9e57db6e44
+6_visualize/tmp/gif_frame_a_20170831_04.png: ea6395d4bce3402e391bd1a815d14376
+6_visualize/tmp/gif_frame_a_20170831_05.png: ab491a27b014593a03ec2dfcb50269fe
+6_visualize/tmp/gif_frame_a_20170831_06.png: bf9e345f85d7d0e29c87764671dfa13c
+6_visualize/tmp/gif_frame_a_20170831_07.png: 3025c43ecc3ab7db900aaa82dc2502dc
+6_visualize/tmp/gif_frame_a_20170831_08.png: e2b264600c7c2906555a1150854e25dd
+6_visualize/tmp/gif_frame_a_20170831_09.png: 171bd48cca2d210e28d841807227bcec
+6_visualize/tmp/gif_frame_a_20170831_10.png: 45c249d706248d90bc162688cf8bda2b
+6_visualize/tmp/gif_frame_a_20170831_11.png: 4ab86ba76c434dcafd68eb855214787b
```
aappling-usgs: wheeee! was this fast?

build/status/Nl92aXpwcmVwL291dC9yaXZlcnNfZnVuLnJkcy5pbmQ.yml
```diff
@@ -0,0 +1,13 @@
...
+type: file
+hash: ca0e1e89f96e3012726138f15d77a357
+time: 2018-07-12 01:14:00 UTC
+depends:
+  1_fetch/out/river_geometry.rds.ind: 0afc7326cf6474c9836b879d206aa545
+fixed: 02a0352cba4af48ec5e70e0ebc354ea1
+code:
+  functions:
+    prep_rivers_fun: dfc3854f0aea696b9e6704849547c2f4
+    write_put_fun: a1f9c3688ca2f519a05753c52f90957e
```
aappling-usgs: can this build/status file be deleted, because rivers_fun.rds.ind is now just a rivers_fun R object?

6_visualize/src/prep_storm_point_fun.R
```diff
@@ -1,11 +1,10 @@
 
-prep_storm_point_fun <- function(ind_file, storm_points_ind_file){
+prep_storm_point_fun <- function(storm_points_ind_file, DateTime){
   storm_points_sf <- readRDS(sc_retrieve(storm_points_ind_file))
-  # parse date from ind_file
-  this_DateTime <- POSIX_from_filename(ind_file)
+  this_DateTime <- as.POSIXct(DateTime, tz = "UTC") # WARNING, IF WE EVER MOVE FROM UTC elsewhere, this will be fragile/bad.
```
aappling-usgs: True. I hope we can and will stick to UTC throughout.

aappling-usgs:
I'm gonna merge because that one little file deletion isn't worth waiting for and we can do it anytime.

6_visualize/log/6_gif_tasks.ind
jread-usgs: Fast enough - seconds per frame, but a lot of frames...

jread-usgs: total build was 10mins

aappling-usgs: Noted. Could be worse, and I'm still hopeful that this multi-file approach will get us the iteration speed we want even if the total runtime is nontrivial.

PR #50, opened 2018-07-12
River Geometry Clean Up

dblodgett-usgs:
demo

There are two layers included in this PR.
1. Major Rivers (dark blue): They are fetched from the [National Water Census Geoserver.](https://cida.usgs.gov/nwc/geoserver/). As of this PR, only order 4 and greater streams are included in this layer. I may want to change that to include all. You can see this in that the light blue lines (from the NLDI) extend upstream of the dark blue lines (from Geoserver) above. The `sf` `data.frame` that the major rivers are in has one row (one geometry) per major river `levelpathid` -- an identifier of the mainstem of a river. These major rivers can be filtered according to the description in [this comment.](https://github.com/USGS-VIZLAB/vizstorm-GIF/issues/11#issuecomment-404534139)
1. Gaged Rivers (light blue): They are fetched from the [NLDI](https://owi.usgs.gov/blog/nldi-intro/) for each site using one upstream mainstem query and another downstream mainstem query. These are then grouped and summarized by gage-id and whether they are up or downstream of that gage. The `sf` `data.frame` that the gaged rivers are in has two rows per gage with a gage id and up or down indicator.

```r
> sf_major_rivers
Simple feature collection with 565 features and 1 field
geometry type:  LINESTRING
dimension:      XY
bbox:           xmin: -526280.3 ymin: -2116672 xmax: 2112163 ymax: -749335
epsg (SRID):    NA
proj4string:    +proj=laea +lat_0=45 +lon_0=-100 +x_0=0 +y_0=0 +a=6370997 +b=6370997 +units=m +no_defs
# A tibble: 565 x 2
   levelpathi                                     geometry
        <dbl>                             <LINESTRING [m]>
 1  250002604 (1611551 -893092, 1629666 -909160.3, 165970…
 2  250002682         (1859823 -1103993, 1867007 -1134663)
 3  250002851 (1953962 -953476.4, 1958439 -964806.7, 1951…
 4  250003088         (1922147 -1030802, 1923230 -1033518)
 5  250003100         (1922764 -1032335, 1923210 -1033151)
 6  250003127         (1922609 -1032338, 1922697 -1033296)
 7  250003173         (1922554 -1030830, 1922487 -1030919)
 8  250003325         (1778875 -1257107, 1781005 -1258853)
 9  250003390 (1859023 -788092.2, 1878104 -812688.5, 1914…
10  250003396       (2014049 -964358.1, 2016609 -966022.4)
# ... with 555 more rows
> sf_gage_rivers
Simple feature collection with 60 features and 2 fields
geometry type:  LINESTRING
dimension:      XY
bbox:           xmin: -547559 ymin: -2116672 xmax: 1779710 ymax: -605856.9
epsg (SRID):    NA
proj4string:    +proj=laea +lat_0=45 +lon_0=-100 +x_0=0 +y_0=0 +a=6370997 +b=6370997 +units=m +no_defs
# A tibble: 60 x 3
   site_id       up_down                          geometry
   <chr>         <chr>                    <LINESTRING [m]>
 1 USGS-02202500 down_main (1742774 -1229149, 1763067 -12…
 2 USGS-02202500 up_main   (1563887 -1111567, 1580772 -11…
 3 USGS-02225000 down_main (1656268 -1277089, 1682511 -12…
 4 USGS-02225000 up_main   (1438053 -1120734, 1465226 -11…
 5 USGS-02397000 down_main (1353776 -1081112, 1341725 -10…
 6 USGS-02397000 up_main   (1441017 -1016731, 1419801 -10…
 7 USGS-02467001 down_main (1138545 -1304116, 1124921 -13…
 8 USGS-02467001 up_main   (1069322 -1129384, 1056561 -11…
 9 USGS-02479000 down_main (1078662 -1484262, 1076613 -14…
10 USGS-02479000 up_main   (1000677 -1345599, 995481.7 -1…
# ... with 50 more rows
```


1_fetch/src/fetch_river_geometry.R
```diff
@@ -1,16 +1,17 @@
-fetch_river_geoms <- function(ind_file, view_polygon, sites, streamorder) {
+fetch_river_geoms <- function(ind_file, view_polygon, sites_ind) {
 
-  sites <- readRDS(sc_retrieve(sites))
+  sites <- readRDS(sc_retrieve(sites_ind))
 
   if(nrow(sites) > 30) {
     set.seed(42)
     sites <- sites[sample(seq_len(nrow(sites)), 30),]
   }
```
jzwart: This is just a temporary fix so we don't select so many rivers?

1_fetch/src/fetch_river_geometry.R
```diff
@@ -1,16 +1,17 @@
...
 
-  bbox <- sf::st_bbox(sf::st_transform(view_polygon, 4326))
+  bbox <- st_bbox(st_transform(view_polygon, 4326))
 
   postURL <- "https://cida.usgs.gov/nwc/geoserver/nhdplus/ows"
 
+  streamorder <- 4 # getting more than we need but not all!
```
jzwart: Do we want to put options like this in the viz_config.yml? I'm thinking of the case where a storm covers a smaller geographic area and maybe we want to have the same density of streams in the viewing area, but this will be dependent on the stream order.

jzwart: I suppose this is just major rivers and we'd still grab smaller streams with the site list

2_process.yml
```diff
@@ -91,3 +92,14 @@ targets:
...
     command: gd_get('2_process/out/storm_points_interp.rds.ind')
+
+  2_process/out/river_geoms.rds.ind:
+    command: process_river_geoms(
+      ind_file=target_name,
+      river_geoms_ind='1_fetch/out/river_geometry.rds.ind',
+      view_polygon=view_polygon,
+      coastal_threshold_sqkm=1000,
+      inland_threshold_sqkm=40000,
+      simplification_tolerance_m=5000)
```
jzwart: I think these thresholds and tolerances should go in the `viz_config.yml` where it will be easy to manipulate based based on the geographic area of the storm

jzwart:
awesome work!

2_process.yml
aappling-usgs: :+1: I like that distinction between configurable stuff that we'd want to build into the design for all vizzies (which wouldn't go in viz_config.yml) and configurable stuff that would vary by event (which would go in viz_config.yml). And I agree that these thresholds would make sense in that second category, though will defer to you @dblodgett-usgs if you disagree.

1_fetch/src/fetch_river_geometry.R
dblodgett-usgs: Correct. I hit the NLDI for each site and having a long list becomes problematic.

1_fetch/src/fetch_river_geometry.R
dblodgett-usgs: I want to change the handling of this. It is a pre-filter so we don't download way too much data, the later filters on drainage area go to more like streamorder 5 or 6. There's a chance it could be removed as downloading a ton of data isn't really a blocker. 

Note the difference between the upstream extent of the light and dark blue lines. That is a result of this filter only grabbing 4th order and higher streams for the dark blue lines. If I switch this to download all major rivers, we would be able to extend the major rivers all the way up stream if we wanted to.

2_process.yml
dblodgett-usgs: This is in this location based on my naiveté only. I think that they should be in viz_config.yml. Will move.

dblodgett-usgs:
demo
Latest. 

We now download ALL the NHDPlus data for the view_polygon and that is split off from the flowlines for sites. At the end of the day we may not use both of these, but options are good. Notice that the dark blue lines now extend all the way to the top of their headwater.

viz_config.yml
```diff
@@ -19,5 +19,10 @@ secondary_geoms:
     xlim: [275, 300]
     ylim: [16, 30]
 
+# Change will cause refetch of entire nhdplus for view_polygon
+fetch_simplification_tolerance_m: 500
```
dblodgett-usgs: I ended up doing this to cut the output file size from the fetcher down. It went from over 300MB to a little over 100MB.

1_fetch/src/fetch_river_geometry.R
```diff
@@ -1,63 +1,81 @@
...
                       '</wfs:GetFeature>')
 
   out <- httr::POST(postURL, body = filterXML)
 
-  sf_major_rivers <- sf::read_sf(rawToChar(out$content)) %>%
-    sf::st_transform(sf::st_crs(view_polygon)) %>%
-    sf::st_simplify(5000)
+  sf_major_rivers <- read_sf(rawToChar(out$content)) %>%
+    st_transform(st_crs(view_polygon)) %>%
+    st_simplify(dTolerance = pre_simplify_m[[1]])
```
aappling-usgs: :+1:

1_fetch/src/fetch_river_geometry.R
```diff
@@ -1,63 +1,81 @@
...
+  saveRDS(sf_major_rivers, data_file)
+  gd_put(remote_ind=ind_file, local_source=data_file, mock_get='none')
+}
+
+fetch_gage_river_geoms <- function(ind_file, view_polygon, sites_ind) {
+
+  sites <- readRDS(sc_retrieve(sites_ind))
+
+  if(nrow(sites) > 30) {
+    set.seed(42)
```
aappling-usgs: :grinning:

1_fetch/src/fetch_river_geometry.R
```diff
@@ -1,63 +1,81 @@
...
+    sites <- sites[sample(seq_len(nrow(sites)), 30),]
+  }
 
   site_list <- paste0("USGS-", sites$site_no)
 
   site_list_DM <- lapply(site_list, navigate_nldi, f_source = "nwissite", mode = "DM")
-  site_list_UP <- lapply(site_list, navigate_nldi, f_source = "nwissite", mode = "UM")
-  names(site_list_DM) <- names(site_list_UP) <- site_list
+  site_list_UM <- lapply(site_list, navigate_nldi, f_source = "nwissite", mode = "UM")
+  names(site_list_DM) <- names(site_list_UM) <- site_list
```
aappling-usgs: just a tip for fun: you can call `lapply(setNames(nm=site_list), ...)` if you want to eliminate the need for this `names()<-` line. or do `site_list <- paste0("USGS-", sites$site_no) %>% setNames(nm=.)`, either way

dblodgett-usgs: yeah... being mentally lazy on this.

2_process.yml
```diff
@@ -22,7 +22,11 @@ sources:
   - 2_process/src/process_precip.R
   - 2_process/src/extract_storm_points.R
   - 2_process/src/interpolate_storm_points.R
+<<<<<<< HEAD
   - 2_process/src/process_river_geoms.R
+=======
+  - 2_process/src/normalize_streamdata.R
+>>>>>>> c7db744994199e0df09616abf49f80c3bb691d6b
```
aappling-usgs: looks like we could keep both of these code lines, delete the git stuff

dblodgett-usgs: woops. missed that.

1_fetch/src/fetch_river_geometry.R
aappling-usgs: no problem, names() works

PR #52, opened 2018-07-12
Normalize stage

jzwart:
Normalizing stream stages for plotting on sparklines.


2_process.yml
```diff
@@ -91,3 +93,10 @@ targets:
       coarse_points_ind='2_process/out/storm_points_coarse.rds.ind')
   2_process/out/storm_points_interp.rds:
     command: gd_get('2_process/out/storm_points_interp.rds.ind')
+
+  2_process/out/normalize_streamdata.rds.ind:
+    command: normalize_streamdata(
+      ind_file = target_name,
+      raw_ind_file = '1_fetch/out/streamdata.rds.ind')
+  2_process/out/normalize_streamdata.rds:
```
aappling-usgs: This is nitpicky, but could we name these output files `normalized_streamdata` with a `d`? To communicate that they're data products rather than functions?

2_process/src/normalize_streamdata.R
```diff
@@ -0,0 +1,15 @@
+# normalize stream stage for sparklines
+
+normalize_streamdata <- function(ind_file, raw_ind_file, gd_config){
+
+  norm_stream <- readRDS(sc_retrieve(raw_ind_file)) %>% # NWIS raw data
+    rename(stage = X_00065_00000, stage_cd = X_00065_00000_cd) %>%
+    select(site_no, dateTime, stage, stage_cd) %>%
+    group_by(site_no) %>%
+    mutate(stage_normalized = (stage - min(stage, na.rm = T)) / (max(stage, na.rm = T) - min(stage, na.rm = T))) %>% # normalizing to min & max ignoring missing data
```
aappling-usgs: :+1:

2_process.yml
jzwart: ah yup, good call

PR #56, opened 2018-07-12
Legend func

jread-usgs:
gif_frame_a_20170826_07

Adding a tiny start of the legend. This is a partial PR and meant to help blodgett w/ the precip percentiles.


1_fetch.yml
```diff
@@ -44,9 +45,16 @@ targets:
     command: viz_config[[I('dates')]]
   stream_params:
     command: yaml.load_file('1_fetch/cfg/stream_params.yml')
+  precip_config:
+    command: viz_config[I(c("precip_cols", "precip_bin"))]
```
dblodgett-usgs: ' " ?!? 😠

6_visualize/src/prep_view_fun.R
```diff
@@ -1,7 +1,7 @@
 
 prep_view_fun <- function(view_polygon){
   plot_fun <- function(){
-    par(omi = c(0,0,0,0), mai = c(0,0,0,0))
+    par(omi = c(0,0,0,0), mai = c(0,0,0,0), bg = '#E1EDF5')
```
dblodgett-usgs: parameterize this? Probably ok to leave hard coded.

1_fetch.yml
jread-usgs: killin' me here

6_visualize/src/prep_view_fun.R
jread-usgs: Yeah, all of the style stuff is currently hard-coded in these functions. If we move this, we'll want to move it all

1_fetch.yml
aappling-usgs: It's so refreshing to work with people even more OCD than i am =)

6_visualize/src/prep_view_fun.R
aappling-usgs: Guideline for this project:
```
if(needs to vary by storm) {
  parameterize it into viz_config.yml
} else if(it'll be the same for all storms) {
  if (it'll be used multiple places) {
    get it into one shared location
  } else {
    leave it where it is
  }
}
```

PR #57, opened 2018-07-12
add loop_tasks and combine_storm_frames()

aappling-usgs:
The loop_tasks() command in 6_visualize.yml should help us with quick iteration on 3 (or some other small number of) frames.

Added a command to combine all the frames into animation_a.gif using imagemagick and gifsicle. These programs will need to be installed and on the right PATH on everyone's computer.

Currently, the animation only combines those frames on which we're iterating in the loop_tasks command. There are notes in 6_visualize.yml on how this should change once we've done enough iteration to firm up the final pipeline.


.gitignore
```diff
@@ -3,11 +3,12 @@
 .RData
 .Ruserdata
 .httr-oauth
+*/tmp/*
```
jread-usgs: :+1:

6_visualize/src/prep_storm_line_fun.R
```diff
@@ -3,7 +3,7 @@ prep_storm_line_fun <- function(storm_line_ind_file){
   storm_line_sf <- readRDS(sc_retrieve(storm_line_ind_file))
 
   plot_fun <- function(){
-    plot(storm_line_sf, add = TRUE, col = 'black', lwd = 2)
+    plot(storm_line_sf[,1], add = TRUE, col = 'black', lwd = 2)
```
jread-usgs: 👍

6_visualize/src/combine_storm_frames.R
```diff
@@ -0,0 +1,12 @@
+combine_storm_frames <- function(gif_file, task_names=NULL) {
+
+  # run imageMagick convert to build a gif
+  if(is.null(task_names)) task_names <- '*'
+  png_files <- paste(sprintf('6_visualize/tmp/gif_frame_%s.png', task_names), collapse=',')
+  magick_command <- sprintf('magick convert -delay 8 -loop 0 %s %s', '6_visualize/tmp/gif_frame_*.png', gif_file)
```
jread-usgs: Might want to have `8` be a %s w/ a decent variable name (`loop_delay`?), since this is a duration of each frame, and we'll want to know what it means and that this is what we should change when we want the gif to be quicker/slower.

6_visualize.yml
```diff
@@ -75,3 +76,31 @@ targets:
       makefile = target_name,
       task_plan = gif_tasks,
       remake_file = I('6_visualize.yml'))
+
+  # Build the timestep-specific functions and png files.
+  # When we like how this is looking, we can set task_names=NULL to build all frames.
```
jread-usgs: "task_names=NULL to build all frames" COOL!

6_visualize/src/combine_storm_frames.R
aappling-usgs: yes, good idea. i'll get that in the next PR

6_visualize/src/prep_storm_line_fun.R
dblodgett-usgs: This does fix the error, but note that we can use the `attr(storm_line_sf, "sf_column")` attribute to get the geometry column

aappling-usgs: :+1:

6_visualize/src/combine_storm_frames.R
aappling-usgs: or `animation_speed`?

jread-usgs: Yes, delay or frame delay is better than what I suggested. That wasn’t a good option

aappling-usgs: https://github.com/USGS-VIZLAB/vizstorm-GIF/pull/61

6_visualize/src/prep_storm_line_fun.R
aappling-usgs: https://github.com/USGS-VIZLAB/vizstorm-GIF/pull/61

Issue #30, opened 2018-07-10
datetime viz function


aappling-usgs:
The date, and possibly also the time (might vary for hurricanes vs inland floods), needs to be displayed on the plot somewhere. This function should provide that plot.

Nice to have: One creative turn to consider might be to make this info graphical in some way - could be a calendar with date squares filling in over time, for example. Something to help people see where they are in time, and connect it to where they were in their own lives at each time point, without having to read and think about a text-based date-time stamp.

jread-usgs:
> could be a calendar with date squares filling in over time, for example
cool idea

jzwart:
closed via #76

PR #58, opened 2018-07-12
Datetime viz fun

jzwart:
function for plotting the datetime on the viz frames. 

Format right now is: abbreviated month day hour:min PM/AM (e.g. Jan 10 12:05 PM)


jzwart:
addresses #30

6_visualize/src/prep_datetime_fun.R
```diff
@@ -0,0 +1,9 @@
+
+prep_datetime_fun <- function(datetime){
```
aappling-usgs: need to get this file into `sources` in `6_visualize.yml`

6_visualize/src/prep_datetime_fun.R
```diff
@@ -0,0 +1,9 @@
...
+  datetime <- strftime(as.POSIXct(datetime), format = '%b %d %I:%M %p')
+
+  plot_fun <- function(){
+    plot(datetime, add = TRUE, col = 'grey')
```
aappling-usgs: Looks like the right format for this file overall, but can you just call plot on a character string? I'm trying it locally and keep getting this error:
```
Error in plot.window(...) : need finite 'ylim' values
In addition: Warning messages:
1: In xy.coords(x, y, xlabel, ylabel, log) : NAs introduced by coercion
2: In min(x) : no non-missing arguments to min; returning Inf
3: In max(x) : no non-missing arguments to max; returning -Inf
4: In plot.window(...) : "add" is not a graphical parameter
```

jread-usgs: I think you want `text`, right? And to be able to position it in the right spot. I have something similar in the `prep_legend` if that is helpful.

jzwart:
I think I cleaned up everything from my brain fart, and added image placement option to the viz_config.yml. It is only implemented for datetime right now but other image functions can use same pattern if we're happy with how I set it up in viz_config. 

image

1_fetch.yml
```diff
@@ -48,6 +48,11 @@ targets:
   precip_config:
     command: viz_config[I(c('precip_cols', 'precip_bin'))]
 
+  legend_styles:
+    command: viz_config[I(c('gage_flood_col','gage_norm_col','hurricane_col','storm_name'))]
+
+  hurricane_col:
+    command: viz_config[[I('hurricane_col')]]
```
aappling-usgs: I'm not sure I agree with moving this graphical config info into 1_fetch.yml. Could you make a case for it here?

jzwart: hmm... I don't think I made that change... Are we looking at the same version of this PR ? I think this must have been merged from an earlier PR

jzwart: I can move it to `6_visuzlize.yml` though. I think it'd fit better there

aappling-usgs: interesting...maybe i'm seeing the extra changes because you opened this PR last night. could you close this one and open a new PR from the same branch? might be cleaner.

jread-usgs: I think I made that change

PR #60, opened 2018-07-12
Precip On the Map

dblodgett-usgs:
Got this taken care of. It may be too slow, but I'm not sure that is actually due to loading data or not. 

gif_frame_a_20170901_23


6_visualize.yml
```diff
@@ -56,6 +56,9 @@ targets:
   storm_line_fun:
     command: prep_storm_line_fun("2_process/out/storm_line.rds.ind")
 
+  storm_points_sf:
+    command: fetch_read('2_process/out/storm_points_interp.rds.ind')
```
aappling-usgs: :+1:

6_visualize/src/prep_precip_fun.R
```diff
@@ -0,0 +1,34 @@
...
+
+  breaks <- unique(c(precip_bins$left_break,
+                     precip_bins$right_break))
+  breaks[which(breaks == Inf)] <- 100000 # Just quiets an error
+  breaks <- c(0, 1, breaks[2:length(breaks)]) # Add a tiny break for alpha
+
+  rgb_ramp <- col2rgb(precip_bins$col, alpha = TRUE)
+  rgb_ramp <- cbind(rgb_ramp[,1],rgb_ramp)
+  rgb_ramp["alpha", 1:4] <- c(0, 100, 180, 220) # fade in from 0
+  rgb_ramp["alpha", ] <- rgb_ramp["alpha", ] * 0.8 # decrease all alpha a bit
```
dblodgett-usgs: Thought it would be nice to have 0 fade out -- just an idea. I think it looks good and helps hide the edge of the precip data field.

6_visualize/src/prep_storm_point_fun.R
```diff
@@ -1,6 +1,7 @@
 
-prep_storm_point_fun <- function(storm_points_ind_file, DateTime, hurricane_col){
-  storm_points_sf <- readRDS(sc_retrieve(storm_points_ind_file))
+fetch_read <- function(ind_file) readRDS(sc_retrieve(ind_file))
```
dblodgett-usgs: Theoretically, using this function will allow us to keep an RDS file full of time steps in memory while all the pngs draw?

6_visualize/src/prep_precip_fun.R
aappling-usgs: @jread-usgs had proposed that we should avoid transparency in order to stick to a 256-color palette for gif, but I've gotta say I really like how these transparent points interact with rivers and state boundaries.

6_visualize/src/prep_storm_point_fun.R
aappling-usgs: That should be the case, yes. [remake](https://github.com/richfitz/remake) is backed by [storr](https://github.com/richfitz/storr), which provides "cached in-memory copies" of objects in the storr.

6_visualize/src/create_gif_tasks.R
```diff
@@ -35,7 +35,19 @@ create_gif_tasks <- function(timestep_ind, folders, storm_cfg){
     },
     command = function(task_name, ...){
       cur_task <- dplyr::filter(rename(tasks, tn=task_name), tn==task_name)
-      sprintf("prep_storm_point_fun('2_process/out/storm_points_interp.rds.ind', I('%s'), hurricane_col)", cur_task$timestep) # pass in storm_points.ind as arg to create_gif_tasks??
+      sprintf("prep_storm_point_fun(storm_points_sf, I('%s'), hurricane_col)", cur_task$timestep)
+    }
+  )
+
+  precip_frame <- scipiper::create_task_step( # not sure why this is called "frame".
```
aappling-usgs: Each task builds a single animation frame illustrating the state at a single timestep. Each step within the task (e.g., one call to the step outlined by `precip_frame`) adds a layer (e.g., precip) to that animation frame.

build/status/Ml9wcm9jZXNzL291dC9zdG9ybV9wb2ludHNfaW50ZXJwLnJkcy5pbmQ.yml
```diff
@@ -1,8 +1,8 @@
 version: 0.3.0
 name: 2_process/out/storm_points_interp.rds.ind
 type: file
-hash: d0966fb9ced4f59227bbe940cba07d77
-time: 2018-07-12 22:24:09 UTC
+hash: e8e0bb137b30af640d5022e6affb5a3c
```
aappling-usgs: I'll be interested to check out the git history of this file once we've accumulated a few more data points. I suspect we're getting OS-specific variations in this file.

6_visualize/src/prep_precip_fun.R
dblodgett-usgs: I guess I though that transparency within layers that go to PNG would work but transparency of the PNG it's self wouldn't?

6_visualize/src/create_gif_tasks.R
aappling-usgs: However, I also agree that "frame" is maybe misleading given that precip is just one layer within the frame. Maybe "precip_layer" would be more informative? or maybe just "precip" would be sufficient?

6_visualize/src/prep_precip_fun.R
aappling-usgs: So -- it _does_ work, and I'm predicting that @jread-usgs will be on board with this. But the introduction of transparency means we can no longer guarantee that the png will have <= 256 colors, which means we have less control over the conversion to 256 colors in the gif, which is why I want jordan to weigh in.

aappling-usgs:
Actually, turns out I can't be that patient. Gonna merge and THEN ask @jread-usgs to weigh in on transparency.

6_visualize/src/prep_precip_fun.R
jread-usgs: It'll work, and I agree that the look is definitely :+1: Nice job!

I would suggest we alter it though and put transparency directly into the color param in the yaml, that way it will show up the same way in the legend as on the map. We can address that later, but that would be my suggestion. 

and yes, with this approach, we'll need to put a bit more trust that the png -> gif conversion is going to keep colors consistent across frames.

6_visualize/src/prep_precip_fun.R
```diff
@@ -0,0 +1,23 @@
...
+
+  colors <- apply(rgb_ramp, 2,
+                  function(x) {
+                    rgb(x[1], x[2], x[3], x[4],
+                        maxColorValue = 255)})
+
+  plot_fun <- function(){
+    plot(precip[[timestep]], add = TRUE, breaks = breaks, col = colors, legend = FALSE)
```
jread-usgs: I think this was supposed to be
```r
plot(precip_rasters[[timestep]], add = TRUE, breaks = breaks, col = colors, legend = FALSE)
```
PR coming w/ the fix

PR #61, opened 2018-07-12
respond to jordan and dave's suggestions

aappling-usgs:
follow-on to PR #57, pulling out the sf geometry better and making animation delay into a viz_config parameter


6_visualize/src/combine_storm_frames.R
```diff
@@ -1,9 +1,10 @@
-combine_storm_frames <- function(gif_file, task_names=NULL) {
+combine_storm_frames <- function(gif_file, animation_cfg=8, task_names=NULL) {
```
jread-usgs: Looks like `animation_cfg` is expected to be a list w/ `$frame_delay_cs`

6_visualize/src/prep_storm_line_fun.R
```diff
@@ -3,7 +3,7 @@ prep_storm_line_fun <- function(storm_line_ind_file){
   storm_line_sf <- readRDS(sc_retrieve(storm_line_ind_file))
 
   plot_fun <- function(){
-    plot(storm_line_sf[,1], add = TRUE, col = 'black', lwd = 2)
+    plot(sf::st_geometry(storm_line_sf), add = TRUE, col = 'black', lwd = 2)
```
jread-usgs: :+1:

6_visualize/src/combine_storm_frames.R
aappling-usgs: good point - default no longer works

PR #63, opened 2018-07-13
remove outdated & unnecessary default arg

aappling-usgs:
Itty bitty one-line change, following up some more on PR #57.


aappling-usgs:
This is so small that I'm gonna merge this myself to avoid wasting a reviewer's time.

PR #65, opened 2018-07-13
minor bug fix

jread-usgs:
needed to generate the pngs


jread-usgs:
typo. Merging my own here

PR #66, opened 2018-07-13
Explanatory animation frames

aappling-usgs:
I've added the framework that should make it pretty straightforward to add explanatory illustration of gage points moving over to their respective sparklines. There aren't actually any graphics in the explanatory phase of the animation yet, but you can see that there are 3 frames in the following animation where there is no precip and no moving storm, just a basemap on which we could draw the gage-sparkline connection.

animation_a


.gitignore
```diff
@@ -12,3 +12,4 @@ lib/cfg/gd_config.yml
 *.rds
 *.feather
 .DS_Store
+*_tasks.yml
```
jread-usgs: :+1:
Yes, was unruly to have that checked in

6_visualize.yml
```diff
@@ -72,45 +67,90 @@ targets:
...
+  # If you're trying to iterate on just a few png frames, delete 6_visualize/log/6_intro_gif_tasks.ind,
+  # then you can call scmake('6_visualize/log/6_intro_gif_tasks.ind') repeatedly.
+  # Once the .ind file exists, you'll need to call scmake('6_visualize/log/6_intro_gif_tasks.ind', force=TRUE)
+  #   to iterate on all frames at once.
+  trial_intro_frames:
+    command: c(I(c(
+      'a_00_001',
+      'a_00_005',
+      'a_00_009')))
+  6_visualize/log/6_intro_gif_tasks.ind:
```
jread-usgs: nice. Loop tasks to .ind :+1:

6_visualize.yml
```diff
@@ -72,45 +67,90 @@ targets:
...
   # When we like how this is looking, we can set task_names=NULL to build all frames.
-  # If you're trying to iterate on just a few png frames, delete 6_visualize/log/6_gif_tasks.ind,
-  # then you can call scmake('6_visualize/log/6_gif_tasks.ind') repeatedly.
-  # Once the .ind file exists, you'll need to call scmake('6_visualize/log/6_gif_tasks.ind', force=TRUE)
+  # If you're trying to iterate on just a few png frames, delete 6_visualize/log/6_storm_gif_tasks.ind,
+  # then you can call scmake('6_visualize/log/6_storm_gif_tasks.ind') repeatedly.
+  # Once the .ind file exists, you'll need to call scmake('6_visualize/log/6_storm_gif_tasks.ind', force=TRUE)
   #   to iterate on all frames at once.
-  trial_frames:
+  trial_storm_frames:
```
jread-usgs: Good. I like this pattern. We could duplicate the pattern to display an interesting "thumbnail" frame for the GIF before it is played (which might show the middle of the storm, which is more interesting than the blank map). 

so:
first frame is 0 delay (there is some way to do this but I haven't figured it out) of a chosen timestep
next n frames are "trial/explanation" frames
next k frames are the storm animated through time. 

Seems like your pattern would support more tack on options like that, or a final explanatory set of frames too, as long as that is handled in the GIF building part.

6_visualize/src/combine_animation_frames.R
```diff
@@ -1,10 +1,10 @@
-combine_storm_frames <- function(gif_file, animation_cfg, task_names=NULL) {
+combine_animation_frames <- function(gif_file, animation_cfg, task_names=NULL) {
 
   # run imageMagick convert to build a gif
   if(is.null(task_names)) task_names <- '*'
-  png_files <- paste(sprintf('6_visualize/tmp/gif_frame_%s.png', task_names), collapse=',')
+  png_files <- paste(sprintf('6_visualize/tmp/gif_frame_%s.png', task_names), collapse=' ')
   magick_command <- sprintf('magick convert -delay %d -loop 0 %s %s',
-                            animation_cfg$frame_delay_cs, '6_visualize/tmp/gif_frame_*.png', gif_file)
+                            animation_cfg$frame_delay_cs, png_files, gif_file)
```
jread-usgs: :+1:
I don't see a better way of specifying *all* pngs that were generated that wouldn't rely on a dir and file name pattern like this. It feels a little different than how explicit we are elsewhere, but gotta do it.

6_visualize/src/create_gif_makefile.R
```diff
@@ -1,10 +1,22 @@
-create_gif_makefile <- function(makefile, task_plan, remake_file) {
+create_intro_gif_makefile <- function(makefile, task_plan, remake_file) {
   scipiper::create_task_makefile(
     makefile=makefile, task_plan=task_plan,
     include=remake_file,
     packages=c('dplyr', 'scipiper'),
     sources=c(
-      '6_visualize/src/create_storm_frame.R'),
+      '6_visualize/src/prep_gage2spark_fun.R',
```
jread-usgs: This is good, but makes me realize we put the .R for the other time-specific plot function builders in `sources` for `6_visualize.yml`. It works because they are connected, but probably better to put those two in here too.

6_visualize/src/create_gif_tasks.R
```diff
@@ -1,9 +1,69 @@
...
 
-create_gif_tasks <- function(timestep_ind, folders, storm_cfg){
+  # aspect/resolution configuration placeholder. see same code in
+  # create_storm_gif_tasks - we should extract this into shared location once we
+  # get multiple configurations really going
+  cfgs <- c('a')
 
+  tasks <- tidyr::crossing(timestep, cfgs) %>%
+    unite(task_name, cfgs, timestep, sep = '_', remove = F) %>%
+    mutate(date_hour = sprintf('00_%03d', timestep),
```
jread-usgs: what is the significance of the "00_03" here?

viz_config.yml
```diff
@@ -12,6 +12,7 @@ precip_bin: 4 # inches per bin
 gage_flood_col: "#175676"
 gage_norm_col: "#4BA3C3"
 hurricane_col: "orange"
+n_intro_frames: 20 # number of frames in the intro animation linking gages to sparklines
 frame_delay_cs: 8 # delay between frames of the gif animation. cs = centiseconds = hundredths of a second
```
jread-usgs: ...we may want a config var for the frame_delay for the intro frames...
:+1: to keep them the same for now, but just noting that.

6_visualize.yml
aappling-usgs: yeah, totally!

6_visualize/src/create_gif_makefile.R
aappling-usgs: agreed, it think it'd be more rigorous to move those into the sources list for 6_storm_gif_tasks.yml

6_visualize/src/create_gif_tasks.R
aappling-usgs: `00_` is just a prefix that will sort before the datestamps of the main fig. we may want to make this `10` or `01` to leave room for a t0 thumbnail gif in the future.

`%03d` is just telling sprintf to plot the integer value of timestep with a total of 3 digits, 0-padding as needed.

viz_config.yml
aappling-usgs: i'm open to that

Issue #12, opened 2018-07-09
sparklines viz function


aappling-usgs:

lindsaycarr:
Handling gaps in the data!

- interpolate small gaps. if the gap is small, linear is likely appropriate. if the gap is larger, interpolation would be a bad idea. Threshold for "small gap" is set in viz_config, but not sure we'd actually change across vizzies
- if the gage goes down for > threshold, we do an X and then don't show data.
- if a gage goes down and then comes back up, we do an X at the end of the first period, a blank spot while it's down, and then pick up with a new polygon when it comes back. Like so: 

![image](https://user-images.githubusercontent.com/13220910/42645569-ffef03ac-85c3-11e8-8789-8223af61bbf6.png)

To start, 5 hrs is going to be the threshold for small.

PR #74, opened 2018-07-13
Plot sparklines for each timestep

lindsaycarr:
Filters gages to just 20. Adds sparklines into pipeline for each timestep png. **Does not** add background stage hydrographs that will exist in a lighter color. Handling gaps in gage data will come at a later date (that's why one of these looks like a crazy X).

image


6_visualize/src/prep_spark_line_fun.R
```diff
@@ -0,0 +1,71 @@
...
+    for(i in sites) {
+
+      # Filter data to just one site & create polygon out of it
+      storm_data_i <- filter(this_spark, site_no == i) %>% arrange(dateTime)
+      storm_data_site_all <- filter(storm_data, site_no == i)
+
+      if(nrow(storm_data_i) == 0) { next } # skip if there's no data at or before this timestep
+
+      flood_stage_va <- as.numeric(unique(storm_data_i$flood_stage_normalized))
+      lowest_stage <- min(storm_data_site_all$stage_normalized, na.rm = TRUE)
```
aappling-usgs: shouldn't lowest_stage always be 0 for normalized stage? to check, i ran
```r
> storm_data %>% group_by(site_no) %>% summarize(min=min(stage_normalized, na.rm=TRUE), max=max(stage_normalized, na.rm=TRUE))
```
and got
```
# A tibble: 20 x 3
   site_no    min   max
   <chr>    <dbl> <dbl>
 1 08033500    0.    1.
 2 08040600    0.    1.
 3 08041000    0.    1.
 4 08041780    0.    1.
 5 08066250    0.    1.
 6 08066500    0.    1.
 7 08067000    0.    1.
 8 08072050    0.    1.
 9 08106350    0.    1.
10 08106500    0.    1.
11 08111500    0.    1.
12 08114000    0.    1.
13 08116650    0.    1.
14 08159500    0.    1.
15 08160400    0.    1.
16 08161000    0.    1.
17 08162000    0.    1.
18 08162500    0.    1.
19 08173900    0.    1.
20 08175800    0.    1.
```

lindsaycarr: I didn't think we wanted to hard code anything in case we changed how we normalized them?

lindsaycarr:
Forgot to say that this is related to #12

6_visualize/src/prep_spark_line_fun.R
aappling-usgs: I think we can and should assume normalized means between 0 and 1 - it adds a lot of code to this function to avoid that assumption, and it's one we have control over in the 2_process phase. You could add a comment to 2_process/src/normalize_streamdata.R making that connection explicit, if you are worried about it.

lindsaycarr: Ok, I see what you mean. I can do that!

6_visualize/src/prep_spark_line_fun.R
```diff
@@ -0,0 +1,71 @@
...
+    y_pos <- y_start # initialize position of first spark
+    for(i in sites) {
+
+      # Filter data to just one site & create polygon out of it
+      storm_data_i <- filter(this_spark, site_no == i) %>% arrange(dateTime)
+      storm_data_site_all <- filter(storm_data, site_no == i)
+
+      if(nrow(storm_data_i) == 0) { next } # skip if there's no data at or before this timestep
+
+      flood_stage_va <- as.numeric(unique(storm_data_i$flood_stage_normalized))
```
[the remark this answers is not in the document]
lindsaycarr: Oops it wasn't when I first started this. Yes, I can remove that.

6_visualize/src/prep_spark_line_fun.R
```diff
@@ -0,0 +1,71 @@
...
+      full_polygon <- data.frame(dateTime = head(storm_data_i$dateTime, 1),
+                                 stage_normalized = min(storm_data_i$stage_normalized,1)) %>%
+        bind_rows(select(storm_data_i, dateTime, stage_normalized)) %>%
+        bind_rows(data.frame(dateTime = tail(storm_data_i$dateTime, 1),
+                             stage_normalized = min(storm_data_i$stage_normalized)))
+
+      # Fill values lower than flood stage with the stage & then create a polygon out of it
+      above_flood_data <- mutate(storm_data_i, stage_normalized = ifelse(stage_normalized < flood_stage_va,
+                                                                         flood_stage_va,
+                                                                         stage_normalized))
```
aappling-usgs: this `ifelse` statement is equivalent to `pmax(stage_normalized, flood_stage_va)`, in case you like that better. (i do but don't feel strongly about it)

6_visualize/src/prep_spark_line_fun.R
```diff
@@ -0,0 +1,71 @@
...
+      above_flood_polygon <- data.frame(dateTime = head(above_flood_data$dateTime, 1),
+                                        stage_normalized = head(above_flood_data$stage_normalized, 1)) %>%
```
aappling-usgs: this is the lower left corner of the polygon, right? i think `head(above_flood_data$stage_normalized, 1)` is redundant with the first row of above_flood_data (bound on the next line), so maybe this line should be `stage_normalized = min(above_flood_data$stage_normalized)` instead?

aappling-usgs: or even simpler: `stage_normalized = flood_stage_va`

6_visualize/src/prep_spark_line_fun.R
```diff
@@ -0,0 +1,71 @@
...
+      storm_data_site_all <- filter(storm_data, site_no == i)
+
+      if(nrow(storm_data_i) == 0) { next } # skip if there's no data at or before this timestep
+
+      flood_stage_va <- as.numeric(unique(storm_data_i$flood_stage_normalized))
+      lowest_stage <- min(storm_data_site_all$stage_normalized, na.rm = TRUE)
+      highest_stage <- max(storm_data_site_all$stage_normalized, na.rm = TRUE)
+
+      full_polygon <- data.frame(dateTime = head(storm_data_i$dateTime, 1),
+                                 stage_normalized = min(storm_data_i$stage_normalized,1)) %>%
```
aappling-usgs: this could be simplified to `stage_normalized = 0` once we're willing to assume that the min is always 0

6_visualize/src/prep_spark_line_fun.R
lindsaycarr: I actually just realized this should not have been working perfectly to begin with. The stage is normalized between 0 and 1, but the flood stage could be above that.
```
> l <- readRDS("2_process/out/normalized_streamdata.rds")
> summary(l$stage_normalized)
   Min. 1st Qu.  Median    Mean 3rd Qu.    Max.    NA's 
0.00000 0.07856 0.52283 0.49679 0.91690 1.00000     224 
> summary(l$flood_stage_normalized)
   Min. 1st Qu.  Median    Mean 3rd Qu.    Max. 
 0.1233  0.4785  0.6223  0.6577  0.7756  1.2874 
```
I could change the limits to `ylim=c(0, flood_stage_va)` ?

6_visualize/src/prep_spark_line_fun.R
```diff
@@ -0,0 +1,71 @@
...
+      # setup a plotting device
+      plot(x=NA, y=NA, type='n', axes=FALSE, xlab="", ylab="",
+           xlim = c(as.POSIXct(viz_config$dates$start, tz = "UTC"), as.POSIXct(viz_config$dates$end, tz = "UTC")),
+           ylim = c(lowest_stage, highest_stage))
+      # put stage polygons on plot
+      polygon(full_polygon$dateTime, full_polygon$stage_normalized, col = viz_config$gage_norm_col, border=NA)
+      polygon(above_flood_polygon$dateTime, above_flood_polygon$stage_normalized, col = viz_config$gage_flood_col, border=NA)
+      points(full_polygon$dateTime[-nrow(full_polygon)], full_polygon$stage_normalized[-nrow(full_polygon)],
+             col = viz_config$gage_flood_col, type='l', lwd=2.5)
+      par(op) # reset plot parameters
```
aappling-usgs: i see there are two layers of par() setting and resetting here, one within the loop and one outside it. why is this particular call needed - which plot parameters get reset here?

6_visualize/src/prep_spark_line_fun.R
aappling-usgs: ooh, good observation...but i think it's reasonable to set the ylims to the normalized data (0 to 1) rather than to compute them from the theoretical (the flood stage), because some streams might stay well below that and others could go well above it, and we'd still want to be able to see the full variation in each type of stream

lindsaycarr: Hmmm, but then will we ever be able to see the flood stage? Or would it get cut off? It is somehow not right now and I can't seem to wrap my head around why

6_visualize/src/prep_spark_line_fun.R
lindsaycarr: Ooooo I just didn't realize that was a function. Wow - game changer! I do this ifelse construction a lot....thanks!

6_visualize/src/prep_spark_line_fun.R
```diff
@@ -0,0 +1,71 @@
...
+        bind_rows(select(above_flood_data, dateTime, stage_normalized)) %>%
+        bind_rows(data.frame(dateTime = tail(above_flood_data$dateTime, 1),
+                             stage_normalized = min(above_flood_data$stage_normalized)))
+
+      y_coords <- c(y_pos, y_pos + vertical_spacing) # y coords relative to plot box
+      fig.new <- c(grconvertX(x_coords, from="npc", to="ndc"), grconvertY(y_coords, from="npc", to="ndc"))
+
+      op <- par(fig=fig.new, cex=2, new=TRUE, mar=rep(0, 4)) # setup parameters
+      # setup a plotting device
+      plot(x=NA, y=NA, type='n', axes=FALSE, xlab="", ylab="",
```
aappling-usgs: I don't see a place where this plot is getting closed. I don't know a lot about base R graphics, but does that mean that future calls to `plot` (in other functions) will be adding to this device rather than the original device? Could that matter for our margins issue?

6_visualize/src/prep_spark_line_fun.R
aappling-usgs: We would see flood stage if and only if the actual stage ever got as high as flood stage. i think that's enough, isn't it?

What are you seeing that makes you say it's not right now?

6_visualize/src/prep_spark_line_fun.R
```diff
@@ -0,0 +1,71 @@
...
+           xlim = c(as.POSIXct(viz_config$dates$start, tz = "UTC"), as.POSIXct(viz_config$dates$end, tz = "UTC")),
+           ylim = c(lowest_stage, highest_stage))
+      # put stage polygons on plot
+      polygon(full_polygon$dateTime, full_polygon$stage_normalized, col = viz_config$gage_norm_col, border=NA)
+      polygon(above_flood_polygon$dateTime, above_flood_polygon$stage_normalized, col = viz_config$gage_flood_col, border=NA)
+      points(full_polygon$dateTime[-nrow(full_polygon)], full_polygon$stage_normalized[-nrow(full_polygon)],
```
aappling-usgs: Maybe also take off the first point in each polygon (the lower-left point)?

6_visualize/src/prep_spark_line_fun.R
```diff
@@ -0,0 +1,71 @@
...
+             col = viz_config$gage_flood_col, type='l', lwd=2.5)
+      par(op) # reset plot parameters
+
+      y_pos <- y_pos + vertical_spacing # increment spacing for next plot
```
aappling-usgs: Which order are sites getting plotted in? From top to bottom or bottom to top? I forget whether y coordinates in R ever start at the top. 

I suspect it'd be easiest to interpret & modify this if the order were from top to bottom, because I think if and when we want to order the sparklines by some variable, our heads will be in dplyr mode and top to bottom will make most sense.

viz_config.yml
```diff
@@ -14,6 +14,11 @@ gage_norm_col: "#4BA3C3"
 hurricane_col: "orange"
 n_intro_frames: 20 # number of frames in the intro animation linking gages to sparklines
 frame_delay_cs: 8 # delay between frames of the gif animation. cs = centiseconds = hundredths of a second
+sparks:
+  xleft: 0.85 # as percentage of figure width
+  xright: 1
+  ybottom: 0.30 # as percentage of figure height
+  ytop: 1
```
aappling-usgs: cool, and i think we can work this into jake's `image_placement` config element (https://github.com/USGS-VIZLAB/vizstorm-GIF/pull/76/files#diff-0c27760a36bf7df1b67ae0e2f1de32e6R20) in a future PR

6_visualize/src/prep_spark_line_fun.R
lindsaycarr: Hmm I guess I only need to reset once within the loop.

```
plot(NA, NA, xlim=c(0,1), ylim=c(0,1), main="background plot")
orig_par <- par()

fig.new.1 <- c(grconvertX(c(0.25, 0.75), from="npc", to="ndc"),
               grconvertY(c(0.15, 0.40), from="npc", to="ndc"))
par(fig=fig.new.1, cex=.5, new=TRUE, mar=rep(0, 4))
plot(1:10, col="red", main="inset plot 1")
par(orig_par)

fig.new.2 <- c(grconvertX(c(0.25, 0.75), from="npc", to="ndc"),
               grconvertY(c(0.65, 0.90), from="npc", to="ndc"))
par(fig=fig.new.2, cex=.5, new=TRUE, mar=rep(0, 4))
plot(1:10, col="blue", main="inset plot 2")
par(orig_par)

points(seq(0,1,by=0.5), seq(0,1,by=0.5), pch=16, cex=2)
```
image

But you have to reset within each loop, or the inset will start to do this:
```
plot(NA, NA, xlim=c(0,1), ylim=c(0,1), main="background plot")
orig_par <- par()

fig.new.1 <- c(grconvertX(c(0.25, 0.75), from="npc", to="ndc"),
               grconvertY(c(0.15, 0.40), from="npc", to="ndc"))
par(fig=fig.new.1, cex=.5, new=TRUE, mar=rep(0, 4))
plot(1:10, col="red", main="inset plot 1")
# WE DON'T RESET HERE (SO NO RESET IN LOOP)

fig.new.2 <- c(grconvertX(c(0.25, 0.75), from="npc", to="ndc"),
               grconvertY(c(0.65, 0.90), from="npc", to="ndc"))
par(fig=fig.new.2, cex=.5, new=TRUE, mar=rep(0, 4))
plot(1:10, col="blue", main="inset plot 2")
par(orig_par)

points(seq(0,1,by=0.5), seq(0,1,by=0.5), pch=16, cex=2)
```
image

6_visualize/src/prep_spark_line_fun.R
lindsaycarr: OH GOOD GRIEF - DUH! Yep, I got it now :) If a site hits flood stage, then that flood stage is included in the 0 to 1 already :)

aappling-usgs: :smile:

6_visualize/src/prep_spark_line_fun.R
lindsaycarr: ah yes, good catch thanks

6_visualize/src/prep_spark_line_fun.R
lindsaycarr: I think that is what resetting the pars are effectively doing. I don't think the margin issues we are seeing are from from this because I've re-built on a different branch and see the same issue

6_visualize/src/prep_spark_line_fun.R
lindsaycarr: I think we need this or the polygon won't get filled?

aappling-usgs: i was referring only to the `points` line - isn't that one just creating the dark line along the top of each polygon?

6_visualize/src/prep_spark_line_fun.R
aappling-usgs: yep, OK

viz_config.yml
aappling-usgs: (note that his image_placement organization has changed since i wrote that)

6_visualize/src/prep_spark_line_fun.R
lindsaycarr: I did not try to do any particular order at the moment. Not really sure where that belongs? Here I suppose.

lindsaycarr:
With this PR, running `scmake('6_visualize/out/animation_a.gif', force=TRUE)` gives you

image

aappling-usgs:
lovely!

PR #76, opened 2018-07-13
Datetime image function, retry 

jzwart:
Re-opened a cleaner PR for datetime image placement. 

Added image placement option to the viz_config.yml. It is only implemented for datetime right now but other image functions can use same pattern if we're happy with how I set it up in viz_config.

image


viz_config.yml
```diff
@@ -15,6 +15,19 @@ hurricane_col: "orange"
...
+image_placement:
+  watermark:
+    x1: 0.01
+    y1: 0.01
+  datetime: # above watermark for now
+    x1: 0.01
+    y1: 0.2
+  legend:
+    x1:
+    y1:
```
aappling-usgs: i like this structure :+1:

6_visualize.yml
```diff
@@ -22,13 +22,14 @@ sources:
   - 6_visualize/src/prep_precip_fun.R
   - 6_visualize/src/prep_rivers_fun.R
   - 6_visualize/src/prep_legend_fun.R
+  - 6_visualize/src/prep_datetime_fun.R
```
aappling-usgs: It would be slightly more precise if you declared this as a source in `create_storm_gif_makefile` rather than in 6_visualize.yml, because the the targets that will need this script are in 6_storm_gif_tasks.yml.

In the branch I'm working on right now, I've moved around a couple other functions with this in mind, because prep_storm_point_fun and prep_precip_fun both also fall into this category. Right now I have:
```r
create_storm_gif_makefile <- function(makefile, task_plan, remake_file) {
  scipiper::create_task_makefile(
    makefile=makefile, task_plan=task_plan,
    include=remake_file,
    sources=c(
      '6_visualize/src/prep_storm_point_fun.R',
      '6_visualize/src/prep_precip_fun.R',
      '6_visualize/src/create_animation_frame.R'),
    file_extensions=c('feather','ind'),
    ind_complete=TRUE)
}
```
so you could add this line as one more source in that function rather than here. I'll probably get a small merge conflict after you've done this, once I get back to working on my branch, but should be easy for me to resolve.

6_visualize.yml
```diff
@@ -43,6 +44,9 @@ targets:
       tmp=I('6_visualize/tmp'),
       log=I('6_visualize/log'))
 
+  image_placement:
+    command: viz_config[[I('image_placement')]]
```
aappling-usgs: It'd be a good idea to subset this even further and only pass to `datetime_fun()` those parameters strictly needed for datetime. Otherwise every time someone nudges the legend around (for example), they'll also end up rebuilding the datetime step, which would be inefficient.

6_visualize.yml
jzwart: ah yup I'll put it in that yaml

6_visualize.yml
jzwart: OK will do

jzwart: remake doesn't like `command: viz_config[[I('image_placement')]][[I('datetime')]]`, any reason why I can't do that? I get an error
 ```
Error in FUN(X[[i]], ...) : While processing target 'datetime_placement':
    Unknown special function [[
```
should I just create `datetime_placement` in `viz_config.yml` rather than indexing here?

jzwart: I created a new `datetime_placement` in `viz_config.yml` for now and index to that

aappling-usgs: was in meeting. new datetime_placement in viz_config is fine. there might be a purrr command for this sort of thing, but the main issue with the command remake didn't like is that remake only allows one function per command. two calls to `[[` is one call too many for remake. so if you wanted to use a nested config structure and then pull out just the datetime bit, you'd need to have two targets - one would pull out all the image placement, and the next would pull out the datetime-specific params from the image placement object.

but to reiterate, the route you chose is just fine.

Issue #71, opened 2018-07-13
add ocean names function


aappling-usgs:

jzwart:
![image](https://user-images.githubusercontent.com/4825431/42706931-19b9825a-869e-11e8-9d9b-1466bdad6c73.png)

aappling-usgs:
Looking pretty good as of #93, but let's also work with the yaml format just a little bit so it can be
```yml
ocean_name_placement:
 g_of_mex:
   x1: 0.05
   y1: 0.70
   label: 'Gulf of Mexico'
```

jread-usgs:
Maybe even take things one step further w/ the yaml to simplify it:
```yaml
  ocean_name_placement:
    -
      x1: 0.05
      y1: 0.70
      label: 'Gulf of Mexico'
    -
      x1: 0.4
      y1: 0.92
      label: 'Atlantic Ocean'
```
since the label and the variable name (e.g., g_of_mex) are kind of redundant here.

PR #78, opened 2018-07-13
Adding Ocean Names 

jzwart:
function for adding ocean names with name placement defined in `viz_config.yml`

image


jzwart:
addresses #71

6_visualize/src/prep_ocean_name_fun.R
```diff
@@ -0,0 +1,31 @@
...
+    # ocean name placement
+    atlantic_x1 <- coord_space_left + coord_width * ocean_name_placement$atlantic$x1
+    atlantic_y1 <- coord_space_bot + coord_height * ocean_name_placement$atlantic$y1
+    g_of_mex_x1 <- coord_space_left + coord_width * ocean_name_placement$g_of_mex$x1
+    g_of_mex_y1 <- coord_space_bot + coord_height * ocean_name_placement$g_of_mex$y1
+
+    x1_coords <- c(atlantic_x1, g_of_mex_x1)
+    y1_coords <- c(atlantic_y1, g_of_mex_y1)
+
+    labs <- c('Atlantic Ocean', 'Gulf of Mexico')
```
jread-usgs: Have you thought about specifying the label, _and_ the x,y locations in the config? 

It would, for example, be a lot easier that way to just remove atlantic if we zoomed in more (and adjust gulf).

jzwart: yup, can do that easily.

jzwart: image
or if we want to change label: 
image

aappling-usgs: :laughing:

PR #89, opened 2018-07-19
restore the sf-ness of normalized streamdata

aappling-usgs:
Bug fix - I was getting an error in the `storm_sites_fun` at this line:
```
plot(sf::st_geometry(norm_data_sites$geometry), add = TRUE,
         pch = 21, bg = gage_col_config$gage_norm_col, col = NA)
```
where the error was something about st_geometry not wanting its argument to be NULL, which seemed to be because `norm_data_sites` wasn't an sf object (even though it did have a geometry column). This change seems to fix the problem.

### DIFF
--- a/2_process/src/normalize_streamdata.R
+++ b/2_process/src/normalize_streamdata.R
@@ -4,7 +4,7 @@ normalize_streamdata <- function(ind_file, raw_ind_file, sites_ind_file){
 
   streamdata <- readRDS(sc_retrieve(raw_ind_file)) # NWIS raw data
   storm_sites <- readRDS(sc_retrieve(sites_ind_file))
-  storm_data <- left_join(streamdata, storm_sites)
+  storm_data <- left_join(streamdata, storm_sites, by='site_no')
 
   # we assume "normalized" stage is between 0 and 1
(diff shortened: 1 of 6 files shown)


2_process/src/normalize_streamdata.R
```diff
@@ -16,8 +16,9 @@ normalize_streamdata <- function(ind_file, raw_ind_file, sites_ind_file){
     mutate(stage_normalized = (stage - min(stage, na.rm = T)) / (max(stage, na.rm = T) - min(stage, na.rm = T))) %>%
     mutate(flood_stage_normalized = (as.numeric(flood_stage) - min(stage, na.rm = T)) / (max(stage, na.rm = T) - min(stage, na.rm = T))) %>%
     ungroup()
+  norm_stream_sf <- sf::st_as_sf(norm_stream)
```
lindsaycarr: Oh oops, good to know! I thought it the column `geometry` was still an `sf` object so it would be fine. I had been running it like this - weird that it didn't throw any errors at me.

aappling-usgs: Yeah, it occurred to me later last night that my errors might have been because my sf or dplyr packages might be out of date relative to yours...but this seems safer all around anyway

Issue #82, opened 2018-07-13
try a new storm


aappling-usgs:
We should make a new google drive folder for this.

Find out what it takes to launch a new storm from scratch.

Change `fetch_streamorder` to be higher (e.g., 4th order) for faster query.

Irma would make it faster too.

jread-usgs:
So far:

- change storm details in viz_config.yml
- create new shared gdrive folder
- change `folder` lib.yml
- remove `.remake` and `build/status` dirs
- delete contents in `6_visualize/tmp`
- change the sites in `filter_sites_custom` (this seems painfully manual right now)

But currently still running into issues that I don't think should pop up if we have everything connected:
```
* Error in prep_spark_funs_data(storm_data, site_data, timestep_ind, spark_config, : no stage data for dateTime<=2017-09-10 16:00:00 and site=03431091; debug with scmake("6_visualize/tmp/gif_frame_a_20170910_16.png", "6_storm_gif_t...
* Error in     DateTime): no stage data for dateTime<=2017-09-10 16:00:00 and site=03431091; debug with scmake("6_visualize/tmp/gif_frame_a_20170910_16.png", "6_storm_gif_tasks.yml")
```

aappling-usgs:
Can we change "change `folder` lib.yml" to "update gdrive folder ID in lib/cfg/gd_config.yml"? I'd like to delete lib.yml, in fact (#95).

Interesting that you had to delete contents of 6_visualize/tmp - they weren't getting overwritten as needed?

Is it possible to use a really bare-bones version of filter_sites_custom at first, then later add in the parts that are painfully manual once we've had a chance to look at the output for all sites?

jread-usgs:
yeah, I think we can get closer to an automated selection of gages that we can further pare down manually. We've used a buffered area near the storm track in the past, or specified a polygon ring as a vector...

jread-usgs:
I added a manual state filter to the viz_config that includes a vector of states to use. I would suggest we use that instead of all of the states that appear in the view (which can be many). For this Irma re-build, I started w/ FL, GA, and AL. 

It is challenging to iterate on getting the view right, since changing it re-downloads all of the precip and river data. Ideally, I'd want to be able to see a png frame with just the geometries and storm-track (no rivers, precip, gages, or sparks) so we can more quickly converge on image size and location of the viewbox. 

[...]

Issue #99, opened 2018-08-16
Move streamflow data fetch from visualize to fetch?


jread-usgs:
Not sure why this happens in `6_` instead of `1_`, but if we move it to `1_`, we can get rid of sites that don't have data during the period of interest, instead of carrying them all the way along to the visualization step.

jread-usgs:
This seems like a fairly big overhaul to change, but the issue right now is we don't know if a site has data for the period of interest until we download it, and we don't do that until we are prepping the plotting function `prep_spark_funs_data()`

jread-usgs:
Not sure why this
```r
warning(sprintf('no stage data for dateTime<=%s and site=%s', DateTime, site))
      shapes[[site]] <- NULL
      next
``` 
was commented out and replaced w/ a `stop()`

aappling-usgs:
Streamflow data fetch does happen in 1_fetch/src/fetch_streamdata.R and is called in 1_fetch.yml line 112.

aappling-usgs:
If we change the stop() back to a warning() we probably need to add/change some logic in prep_gage2spark_fun() to accommodate an unreliable number of sparklines

jread-usgs:
I think this issue was created because of my confusion as to where these things are happening. The fetch isn't happening in `6_`, it just seemed that way to me in some of the builds based on the order of target builds.

PR #100, opened 2018-08-16
Irma

### LINKED ISSUE #96: fetch_precip_values should use POSIX for dates
It uses `as.Date` and then compares to .nc POSIXs. This also means that when we set an hour for start or end in the config that it gets flattened into the date. Simple fix is 
```r
  start_date <- as.POSIXct(dates$start, tz = 'UTC')
  end_date <- as.POSIXct(dates$end, tz = 'UTC')
```

while I'm in here, why are we excluding the start and end dates themselves? 
```r
t_inds <- which(t_vals > start_date & t_vals < end_date)
```

jread-usgs:
There are various fixes scattered in here. Let me know if I should pull things out and remove the changes that are specific to #82 

fixes #96


6_visualize/src/prep_spark_line_fun.R
```diff
@@ -49,10 +49,11 @@ prep_spark_funs_data <- function(storm_data, site_data, timestep_ind, spark_conf
...
     if(nrow(storm_data_i) == 0) {
-      stop(sprintf('no stage data for dateTime<=%s and site=%s', DateTime, site))
-      # warning(sprintf('no stage data for dateTime<=%s and site=%s', DateTime, site))
-      # shapes[[site]] <- NULL
-      # next
+
+      #stop(sprintf('no stage data for dateTime<=%s and site=%s', DateTime, site))
+      warning(sprintf('no stage data for dateTime<=%s and site=%s', DateTime, site))
+      shapes[[site]] <- NULL
+      next
```
aappling-usgs: Maybe for now...but see https://github.com/USGS-VIZLAB/vizstorm-GIF/issues/99#issuecomment-413666567. Is there some way we could push this warning-based check earlier (into 1_fetch) and let it stay a stop() here? Otherwise I think we need to add more logic to gage2spark so that the site points and sparklines match up and the intro animation doesn't break.

aappling-usgs:
@jread-usgs Could you fix the merge conflicts? I'd like to switch to Irma where hopefully my remake pipeline isn't broken.

jread-usgs:
Closing this in favor of #109

Issue #86, opened 2018-07-13
handle gaps in gage data


lindsaycarr:
See comments in https://github.com/USGS-VIZLAB/vizstorm-GIF/issues/12#issuecomment-404565831 for criteria.

lindsaycarr:
^ Hmmm had not considered that when I originally wrote this. It's written vaguely enough to include both, but I think MVP would be just the spark line

aappling-usgs:
I like the idea of throwing an error if we have more than one gap - i worry about just ignoring the possibility but agree that it could be a pain to code for `n` gaps, so an error seems like a good compromise.

The code changes for gaps in hydrograph vs gage point will probably happen in quite different functions, but I think we need a plan for what to do with the gage point - I could reluctantly settle for leaving the gage unchanged, as @lindsaycarr is suggesting, but I would rather see the change on the map simultaneously. Open to further disagreement/discussion, though.

If changing both sparks and points is too big for one PR, then 2 PRs for this issue would be fine with me, but I'd prefer to see both tasks resolved, and at least today I'm OK with lumping those two tasks into this one GitHub issue rather than breaking them out.

jread-usgs:
Yeah, I agree that I'd prefer both, to the point where we should consider the MVP to just manually filter out gages w/ gaps, but the thing we are shooting for is the Maria experience/styling.

aappling-usgs:
Right now I seem to have things working for the sparklines (still need to do the site points) except I was still working on Harvey and the Drive might be corrupt because now my computer is trying to rebuild precip (which it can't. not a mac.). So I won't put in a PR quite yet, but here's what I could do before that rebuild got triggered:
![image](https://user-images.githubusercontent.com/12039957/44241201-59dfc580-a190-11e8-84b7-8a82158d45fb.png)

jread-usgs:
Looking good! :+1: 
I am stuck on some NWIS 503s, so skipping out for the night. I wonder if it is worth tackling that precip build to work on windows? Or at least capture it in an issue.

aappling-usgs:
It would certainly be useful if anybody could do the precip build. I just added #102 but also categorized it as a Could Do because I think our MVP should be a releasable viz.

aappling-usgs:
Resolved by #106 and #114

PR #106, opened 2018-08-17
Support gaps in sparklines

aappling-usgs:
animation_a

Built on docker! Will make a separate docker PR with just a few changes for that.

This PR only addresses the sparklines, not the symbols of the gages on the maps. It seems complex enough as a full PR that I'd like to get review now and will do gage symbols as a separate PR. It's mainly complex because we ended up rebuilding a bunch of files -- apparently the hash for view_config had changed.


aappling-usgs:
animation_a
:point_up: with most of the intro frames

I'm still skipping lots of frames in the build because it takes so.darn.long to build the full GIF

aappling-usgs:
This PR is for #86 (all but the gage location points)

aappling-usgs:
Jordan asked how graphics compare, Windows cairo vs Linux cairo. Here's Linux:
gif_frame_a_20170830_13
and here's Windows:
gif_frame_a_20170830_07
There are slight differences in positioning of text, and I think Linux looks slightly better, but they're pretty close and both probably beat by Mac Quartz.

And here's a gif with Linux at the beginning and end and Windows in a middle chunk of frames:
animation_a

PR #107, opened 2018-08-17
Docker

aappling-usgs:
Here's what I needed to change to get docker (the aapplingusgs/vizstorm-gif image from Docker Hub) to work. Other than these changes, and the setup to get docker machine working on a given computer, it's really quick to spin this up - just call `docker-compose up` from the terminal and navigate to the RStudio URL in the browser (see the README). And I'm not seeing the slowdown that I was worried about based on early experience last week - I think that may have just been a packrat slowdown, actually.

I'm not sure how to handle the docker-compose.yml. It's convenient to have in the repo, but there's one line that's user-specific because you're mapping a local folder to a docker folder, and I couldn't get relative links to work at least for my Win7 machine. The option I'm trying now is to commit a mostly-there version so we all have something to start with, but then we'll each need to modify the file...and then live with always seeing it as Modified in our Git tab in RStudio? Not awesome, but I can't think of a better way yet. Suggestions welcome.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -15,3 +15,4 @@ lib/cfg/gd_config.yml
 *_tasks.yml
 6_visualize/log/*
 build/status/Nl92aXN1YWxpemUvbG9n*
+docker-compose.yml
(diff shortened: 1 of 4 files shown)


aappling-usgs:
Yeah...just having it sit there Modified isn't going to work - you can't switch branches! I think we'd better just gitignore it but put info in the readme about how to make your own. Coming.

aappling-usgs:
Happy to do that. Merging should not interfere with your ability to keep on keepin' on with your Mac (or to try out Docker when you're ready).

PR #108, opened 2018-08-19
Opendap subset

jread-usgs:
This is solely for #109
Instead of reading the .nc from a URL, we subset it to a temporary file w/ `geoknife`. From there, we do the steps in `precip_spatial` and `precip_values` from that temp file, and output a list that contains both. I modified a function in `2_` to accept the list instead of the two separate files. 

I used the list because saving and posting the .nc as an intermediate file (which would be required to build the `spatial` and the `values` separately) was too time consuming and the file was too big (~0.5GB for the Harvey time/space domain). Posting that big file to drive was _really_ slow for me from work on Friday (~60 mins), but it was unclear how much the USGS network had to do w/ that. Regardless, I re-wrote some things to avoid that file. The resulting pngs look the same to me. 

Would be great if @dblodgett-usgs could look at this at some point, but I know he is on vacation.


dblodgett-usgs:
Looks fine if it works. I would expect it to, but switching from the full domain to a subset netcdf file has some risks regarding bounding box overlaps. The assumptions made in the GDP and in the code I wrote must jive well.

It's unfortunate we have to do this at all. Given the time we put in, we should approach mdsumner about why the version of ncdf4 he has on github doesn't do what it says it should. I'm guessing we aren't installing it correctly. If we can fix that then an intermediate download with the GDP isn't needed at all.

aappling-usgs:
Looks OK to me, but Dave had the above hesitation about bounding box overlaps, and there are merge conflicts.

jread-usgs:
Let's let this sit for a bit as I don't think it is pressing since Alison has docker going and I don't think others are building. Adding this in is a little bit slower than the URL method, so it would be good to skip it if we can in the end. Also would be good to have others try the install of that github source for `ncdf4` and see if they have luck on windows before we give up on it.

Issue #97, opened 2018-08-16
MEGA issue: handling filtering of gage data


jread-usgs:
Calling this a MEGA, because it isn't really defined as an issue, and may break out into several issues/tasks. 

Currently, we get gages from all sites w/ flood stage that are in states that appear in the view. Next we pass in a vector of sites "to keep". This is very manual right now, and I'm not sure what others have done to select the sites that belong in the site vector. I also don't know if we are filtering out sites that don't have observations during the requested period, because I keep running into `Error in prep_spark_funs_data(storm_data, site_data, timestep_ind, spark_config,  : 
  no stage data for dateTime<=2017-09-10 03:00:00 and site=02246500`

A couple of things I have noticed playing around with the code:
 - we _could_ filter out gages that are in one of the states in the view, but the actual gage location is outside of the view. This will save time from the fairly expensive step of pulling discharge data we don't need
 - we _could_ specify a vector of state codes to look to for the site query (I am doing this in the branch for #82 currently)
 - we _could_ specify a lat/lon coordinate ring (which can be an odd shape as long as the coordinates "closes") and use all sites that fit the other criteria (e.g., has flood_stage) in there, minus any individual sites we manually exclude. 
 - we _could_ filter to sites that exceed flood_stage during the period we are looking at, but that only helps us pare down the number of gages that appear, not the number of gages we pull data for. 
 - others?

jread-usgs:
Another option to add to the pile (and would help us recognize sites w/o any data during the visualization period) is to pull `dv` data first, but use the `00001` statCd (maximum value during the day).

_If_ any of the dv's for a given site are > flood_stage, then we'll use this site (or manually filter it _out_). 

This also assumes we're applying some initial filters, like "within states X,Y,Z", and "drainage area > K", and "has data that overlaps w/ the period of interest"

The dv data fetch should be a lot quicker than the uv fetch, so I think this could help us if we want these GIFs to be "flooding rivers".

wdwatkins:
We could do something based on total rainfall over HUCs, in an effort to capture the gages that would presumably have the biggest response to the rain.  Maybe intersecting HUCs and the 10 highest precip cells or something.  This would require some relatively sophisticated preprocessing before downloading the discharge data.

PR #109, opened 2018-08-19
Matthew

### LINKED ISSUE #82: try a new storm

jread-usgs:
fixes #82 

Thinking about this pattern for different storms - we create a branch from `upstream` named `{storm_name}-{NOAA_storm_key}`, then PR to that w/ storm-specific code. 

The tricky part w/ this pattern is pulling in changes from upstream master that we want to apply to storm branches. I think we probably want to delete all data, .ind files, `.remake` dir, and the `build/status` dir locally (on the storm-specific branch), then pull in changes from upstream master, then delete ind and build/status again. Or is there a better way? Otherwise we are going to have conflicts w/ .inds and build.yamls. 

Creates: 
animation_a

I didn't do much for formatting and text work, and didn't spend any time selecting sites, so we don't have any for FL, but see #97 discussion.


aappling-usgs:
GIF looks good. Are you able to build the intro sequence as well?

Good point that the differences in build/status and .ind files will be an issue among branches. 
> The tricky part w/ this pattern is pulling in changes from upstream master that we want to apply to storm branches. I think we probably want to delete all data, .ind files, .remake dir, and the build/status dir locally (on the storm-specific branch), then pull in changes from upstream master, then delete ind and build/status again. Or is there a better way? Otherwise we are going to have conflicts w/ .inds and build.yamls.

I'm wondering about the deletion of everything twice...does the above approach mean you have to rebuild all the data each time you pull changes from master?

Some options, not mutually exclusive:
1) Build each vizzy branch in a different local folder - that way your .remake, build/status, and data files can all be unique to each storm without needing to overwrite other storms.
2) Delete all storm-specific files from the master branch permanently: build/status/*.yml files, *.ind files, and lib/cfg/gd_config.yml. This would make the master branch impossible to test on its own, but you could test the code in the context of a storm, and there wouldn't be conflicts between master and any storm-specific branches.

I like  `{storm_name}-{NOAA_storm_key}` as the branch naming convention. Could we safely go even simpler to `{storm_name}-{2_digit_year}`, e.g., `matthew-16`? All storm names are unique within a year, right?

6_visualize/src/create_gif_tasks.R
```diff
@@ -39,6 +39,23 @@ create_intro_gif_tasks <- function(intro_config, folders, storm_start_date){
       )
     }
   )
+  legend_frame <- scipiper::create_task_step(
```
aappling-usgs: Huh. Haven't I already been working from code that included this? I thought you had added a version where DateTime was dynamic (for the storm animation) and then I had added a version where DateTime was NA (for the intro animation): https://github.com/USGS-VIZLAB/vizstorm-GIF/pull/106/files#diff-0de26640bfa961faaa7a7265e37434e9R43. Maybe this was a merge issue because line 60 just below starts another bunch of lines [re]defining `legend_frame`.

jread-usgs: weird...perhaps this was related to what was branched from master to create this new branch on `upstream`? Maybe it didn't have a recent PR...which is odd, because I thought I just created this branch on USGS-R.

aappling-usgs: Yeah, and it says it's up to date. Maybe it had to do with overlap between when my PR was open and when yours was?

jread-usgs: odd, I will clean it up.

jread-usgs:
> GIF looks good. Are you able to build the intro sequence as well?

no...and I've had a mix of results with adding the intro. Would you mind sharing your workflow when you add the intro to the build? I end up getting errors related to missing png files lately, but I haven't dug into it.

 > I'm wondering about the deletion of everything twice...does the above approach mean you have to rebuild all the data each time you pull changes from master?

yes, I have to do big rebuilds when I switch branches or pull from master. I don't think this is ideal at all, but was thinking my current use-case is kind of temporary (hopping between storms and testing things out). 

out of your two options above, option two seems to be the easiest to implement in our current set up, and seems like a clever idea to me. Did you have a preference?

jread-usgs:
I don't think we are yet in a place where we know how to spawn a new storm in the repo. I consider this PR to be a test case or an example of one pattern. I think we should probably do something different after reading Alison's comments:
>>>
- Build each vizzy branch in a different local folder - that way your .remake, build/status, and data files can all be unique to each storm without needing to overwrite other storms. 
- Delete all storm-specific files from the master branch permanently: build/status/*.yml files, *.ind files, and lib/cfg/gd_config.yml. This would make the master branch impossible to test on its own, but you could test the code in the context of a storm, and there wouldn't be conflicts between master and any storm-specific branches.

I am going to close this and focus on getting the good fixes in here onto a PR to `master`, and then we'll figure out the pattern of how we build a new storm.

aappling-usgs:
Sounds good - you've made good progress regardless. I think those two options can both be used together, or option 1 can be optional on a person-by-person basis. If you like it, too, then I think we should definitely do option 2.

aappling-usgs:
For rebuilding the intro, I wonder if you need to call one or all of the following:
```
scmake('6_visualize/log/6_storm_gif_tasks.ind', force=TRUE) # shouldn't need force, but might help
scmake('final_intro_frames', force=TRUE) # more likely to need force here
scmake('6_visualize/out/animation_a.gif', force=TRUE)
```
The second line is only needed if you're using final_intro_frames as input to the `combine_animation_frames` command under `6_visualize/out/animation_a.gif`. You may also need to call `scmake('final_intro_frames', force=TRUE)` periodically in the same conditions and for the same reason: neither function explicitly relies on the contents of the /tmp directory, but the results of `existing_png_tasks` describe those contents and so need to be rebuilt whenever the list of actual .png files within /tmp/ changes.

Until you merged my PR this morning, I don't think we had shared working code for the intro frames because we needed to switch to the dynamic version of the legend function. So that may also have played a role in your troubles.

jread-usgs:
I think `scmake('final_intro_frames', force=TRUE)` is what I was missing here. 

What I'd like to do is progress far enough into the "base storm" and then move into option 2 for this master branch. Sound good?

aappling-usgs:
I think that sounds good but am not clear on your precise plan. What's the "base storm" - Matthew, or storm rather than intro animation, or something else? What are your criteria/goals for "progress far enough into" the base storm?

jread-usgs:
I was thinking the base storm is Harvey, which is what the master branch is on. I don't have a precise plan, but it seems there are some open issues that should be wrapped up before we delete all of the .ind and build yamls from the master. I was planning on making some progress there. 

With the instructions above, when do you build from the `6_intro_gif_tasks.yml`? Should that first command be `scmake('6_visualize/log/6_intro_gif_tasks.ind', force = TRUE)` instead? either way, I am getting a `Error in mutate_impl(.data, dots) : 
  Column `y` must be length 1 (the number of rows), not 3` error on the spark plot plotting function when I build the final .ind file from `6_intro_gif_tasks.yml`

aappling-usgs:
Got it. Sounds good - will indeed be easier to test and debug while there are still .ind and build/status files in the repo.

Oops, yes, I should have written `scmake('6_visualize/log/6_intro_gif_tasks.ind', force = TRUE)` above.

Issue #92, opened 2018-08-09
try storm with a different image size (for presentation)


lindsaycarr:
We would like to use this vizstorm gif in the HDAC vizzies presentation. It would be great if it could be a square (so the Instagram aspect ratio) because we could fit the gif and some text onto a wide format slide. This will obviously take some rejiggering of the components that make up the viz.

- [ ] resize the gif to be a square
- [ ] ignore gages that have gaps (there still appears to be a funky one, 7th from bottom)
- [ ] slow the gif down and actually march through the entire time period

Here's the gif created from the most recent PR (#91):

![vizstorm](https://user-images.githubusercontent.com/13220910/43923068-50506704-9be6-11e8-873a-432d88165571.gif)

jread-usgs:
I can probably take this on w/ #82 since they are similar.

jread-usgs:
![gif_frame_a_20170902_00](https://user-images.githubusercontent.com/2349007/44343481-a4dd2f80-a453-11e8-8132-21b507f6d7e3.png)

@lindsaycarr let me know if you would like to use this storm. I can generate the gif and code is [here](https://github.com/jread-usgs/vizstorm-GIF/tree/Harvey-al092017)

lindsaycarr:
@jread-usgs yes, I think that would work nicely!

Issue #69, opened 2018-07-13
final layout refresh


aappling-usgs:

jread-usgs:
A couple things that should probably happen here include positioning of legend and watermarks based on common args. Such as 
```r
pos = c('botleft','upleft','upright','botright')
```
the watermark function is written to be `botleft` right now, and the legend is `botright`. These are good positions for a landscape GIF for certain storms, but we'll need the control to move these things around easily. 

Same issue w/ the `dateTime` function, but that is likely to change with #79

jread-usgs:
Some legend stuff gets funky when we change the aspect of the figure. Consider using `dev.size("px")` from within the legend plot function to move or size certain things according to number of pixels instead of percentage. This would work well for the bin width for the precip, for example.

PR #111, opened 2018-08-19
Layouts

### LINKED ISSUE #96: fetch_precip_values should use POSIX for dates

jread-usgs:
fixes #96 
progress for #92 (w/ adjustable legend and watermark locations)
Added the position stuff in #69, but it is still percentage (not pixel) based. 

I pulled in a number of small fixes that were in the Irma/Matthew branches but should be in this base storm (Harvey) master branch. I rebuilt everything with those changes (see png):

gif_frame_a_20170831_11


6_visualize/src/create_animation_frame.R
```diff
@@ -1,7 +1,7 @@
 create_animation_frame <- function(png_file, config, ...) {
 
   # open the plotting device. this should eventually be a gif function, but i don't know how that looks yet
-  png(filename=png_file, width=config$width, height=config$height, units='px', type='cairo')
+  png(filename=png_file, width=config$width, height=config$height, units='px')
```
aappling-usgs: Could we stick a switch/ifelse statement in here to support a best-available `type` for each OS?

PR #114, opened 2018-08-21
Site symbols for gaps in data

aappling-usgs:
This PR makes the site dot be an X during periods when data are unavailable for that dot.

I'm rebuilding the animation now, but because the one site with a data gap has been removed, this will actually only show that my code doesn't break on sites without gaps. I'll share that animation and then rebuild a single png to illustrate what this PR is meant to do.


aappling-usgs:
animation_a

aappling-usgs:
Adding back in some bad sites, including 08161000 which "plots funny" according to the comments in filter_sites_custom. Turns out the reason it was plotting funny was because that site got duplicated somehow in the list during filter_sites_geom, so I've resolved that now.

Before site de-duplication:
gif_frame_a_20170831_19 2

After:
gif_frame_a_20170831_19

I also added some code to make it less likely that site points will temporarily drop off the map (this was happening when the NWIS timeseries flat-out skipped time points rather than including NAs for those points). Will add commits for that and deal with the merge conflict shortly.

aappling-usgs:
Resolves #86 and now really truly looks like what I intended. 
animation_a

Issue #95, opened 2018-08-15
remove lib.yml


aappling-usgs:
Seems to be more trouble than it's worth, because all it does is help create lib/cfg/gd_config.yml if that 1-line file doesn't yet exist. And if you update lib.yml after you already have a lib/cfg/gd_config.yml file, the change won't get propagated to gd_config.yml even though that's where we most need the Drive ID.

aappling-usgs:
Resolved by #117

PR #117, opened 2018-08-22
replace lib.yml with README info

aappling-usgs:
Added notes in the README about how I think this project will/should work soon. For the moment we still have all those test-storm-specific files in master, which I think we should keep for another couple of days.


aappling-usgs:
Resolves #95

PR #123, opened 2018-08-24
add waterbodies

### LINKED ISSUE #119: add water bodies
Showing water bodies would be useful for orientation in the August 2018 flooding near Madison, and for Lake Pontchartrain to cover up what looked like a flow line through the ocean

image

dblodgett-usgs:
Fixes #119 -- I haven't been able to get the precip to build with repeated network failures. But this builds through process and it should wire up right in plotting. A little copy paste in the fetch function that could have been avoided, but not sure it's worth refactoring other code at this point.

### DIFF
--- a/1_fetch.yml
+++ b/1_fetch.yml
@@ -130,18 +130,29 @@ targets:
     command: viz_config[I('fetch_streamorder')]
 
   1_fetch/out/major_river_geometry.rds.ind:
-    command: fetch_major_river_geoms(ind_fil=target_name,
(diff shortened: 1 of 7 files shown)


aappling-usgs:
Cool! This is excellent, Dave, thanks. Like you, I'm hitting network failures on fetching precip, so I haven't been able to build a viz yet today, so I'll do some debugging of that bit this afternoon or next week (and maybe the network will get better). Looking forward to seeing the water bodies in there!

Issue #101, opened 2018-08-17
Dynamic legend fails poorly when there is no longer a dot from the storm


jread-usgs:
Sometimes we'll want the animation to extend past when NOAA has a storm track/dot for a hurricane. The legend currently throws an error when we hit a timestep where this is the case (see [oct 10](https://owi.usgs.gov/vizlab/hurricane-matthew/) for example. 

We already know we want to be able to support a gif w/o a hurricane, but what do we do when the hurricane/TS is only there for part of the animation?

jread-usgs:
Not sure what should happen when this is the case:
 - show the color/category of the last dot that appeared?
 - remove this part of the legend for timesteps when the storm is no longer active?
 - change to a generic storm name (w/o category)?

aappling-usgs:
Unless we have the storm name somewhere else on the GIF, I think it'd be nice to have some variant stay written in the legend. What if we used the storm track line as the symbol (line in place of dot) and switched to the generic storm name?

jread-usgs:
I prefer that proposal ☝️ 
> used the storm track line as the symbol (line in place of dot) and switched to the generic storm name

jread-usgs:
Fixed

PR #128, opened 2018-09-05
Legend for no storm point

lindsaycarr:
Keep line in legend and storm name when there is no storm point at a particular time step (#101, but doesn't address absence of storm track). 

Also, adjust all legend points to be centered with their text.

image

### DIFF
--- a/6_visualize/src/prep_legend_fun.R
+++ b/6_visualize/src/prep_legend_fun.R
@@ -11,12 +11,13 @@ prep_legend_fun <- function(precip_bins, legend_styles, timesteps_ind, storm_poi
   }
   this_DateTime <- as.POSIXct(DateTime, tz = "UTC") # WARNING, IF WE EVER MOVE FROM UTC elsewhere, this will be fragile/bad.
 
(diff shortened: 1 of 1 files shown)


6_visualize/src/prep_legend_fun.R
```diff
@@ -41,6 +42,8 @@ prep_legend_fun <- function(precip_bins, legend_styles, timesteps_ind, storm_poi
     ybottom <- coord_space[3]
     dot_x <- x_edge+bin_w/2*shift_dir
     dot_txt_x <- x_edge+bin_w*0.7*shift_dir
+    seg_x <- x_edge+bin_w/3*shift_dir
+    center_to_txt_y <- strheight("A")/3 # height of character divided by three seems to do the trick
```
aappling-usgs: agreed - this is a bit magical, especially since the text elements below add a cex=1.5 that would make strheight not exactly accurate, but it does look good, so let's keep it until the time if and when we need something more directly tied to text height (which may never happen)

PR #129, opened 2018-09-05
Color changes

aappling-usgs:
Mainly adjusting colors of water bodies, with a grey-green for marshes. Also lightened the storm track - what do you think?
gif_frame_a_20170825_00

### DIFF
--- a/1_fetch.yml
+++ b/1_fetch.yml
@@ -54,7 +54,7 @@ targets:
     command: viz_config[I(c('precip_cols', 'precip_bin'))]
 
   legend_styles:
-    command: viz_config[I(c('gage_flood_col','gage_norm_col','hurricane_cols','hurricane_col_names','storm_name'))]
+    command: viz_config[I(c('gage_flood_col','gage_norm_col','hurricane_cols','hurricane_col_names','storm_name','storm_line_col'))]
 
   hurricane_cols:
     command: viz_config[[I('hurricane_cols')]]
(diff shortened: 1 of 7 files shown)


aappling-usgs:
@lindsaycarr To use these changes in Gordon, you'll need to add these lines to your viz_config.yml:
```
major_river_col: "#1874cd"
gage_river_col: "#5d9ddc"
lake_col: "#a2c7eb"
marsh_col: "#afd0c7"
ocean_col: "#e7f1fa"
storm_line_col: "grey70"
```

lindsaycarr:
I'm kind of wondering if it's too close to the precip 8-12 or 12-16 categories

aappling-usgs:
The marsh green? Me too, but unless we go for something glaringly bright, I haven't been able to find a color that would do better while still being a little greenish. Another option would be to mess with the precip color scale... I also looked into a pattern fill, ideally something like this:
image
or simpler, but I haven't been able to figure out how to do any sort of pattern yet, especially with sf plotting...

aappling-usgs:
Maybe something grayer?

gif_frame_a_20170825_00

lindsaycarr:
I think the greyer bit is nicer. Overall I like this scheme, I just think we should be open to changing the color when we see it in full majesty once a completed gif is shown

aappling-usgs:
OK, let's do it. You can experiment with different colors by editing the storm-specific viz_config.yml anyway.

PR #130, opened 2018-09-05
keep UTC when constructing dateseq for precip fetch

aappling-usgs:
I think this should do it. Haven't tested yet, though. Fixing the issue that the c() in the old code was causing dateseq to be in the user's local time rather than in UTC, which meant we were failing to catch some precip timestamps at the end of the date range. Will test next, but since it takes tens of minutes, figured I could also share.

### DIFF
--- a/1_fetch/src/fetch_precip_nc.R
+++ b/1_fetch/src/fetch_precip_nc.R
@@ -20,7 +20,7 @@ fetch_precip_data <- function(ind_file, view_polygon, times) {
   stencil_pts <- as.stencil_pt('xmin', 'ymin', 'low_left') %>% cbind(stencil_pts)
 
   # break times into chunks to avoid WAF rejections of queries that are "too big"
-  try_formats = c("%Y-%m-%d %H:%M:%OS", "%Y/%m/%d %H:%M:%OS", "%Y-%m-%d %H:%M", "%Y/%m/%d %H:%M", "%Y-%m-%d", "%Y/%m/%d")
+  try_formats <- c("%Y-%m-%d %H:%M:%OS", "%Y/%m/%d %H:%M:%OS", "%Y-%m-%d %H:%M", "%Y/%m/%d %H:%M", "%Y-%m-%d", "%Y/%m/%d")
   dateformat <- try_formats[sapply(try_formats, function(fmt) all(!is.na(as.POSIXct(unlist(times), format=fmt))))]
   dateseq <- as.POSIXct(unique(c(
     format(seq(as.POSIXct(times$start, tz='UTC'), as.POSIXct(times$end, tz='UTC'), by=as.difftime(3, units='days')), dateformat),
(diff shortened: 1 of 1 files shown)


1_fetch/src/fetch_precip_nc.R
```diff
@@ -20,10 +20,12 @@ fetch_precip_data <- function(ind_file, view_polygon, times) {
   stencil_pts <- as.stencil_pt('xmin', 'ymin', 'low_left') %>% cbind(stencil_pts)
 
   # break times into chunks to avoid WAF rejections of queries that are "too big"
-  dateseq <- unique(c(
-    seq(as.POSIXct(times$start, tz='UTC'), as.POSIXct(times$end, tz='UTC'), by=as.difftime(3, units='days')),
-    as.POSIXct(times$end, tz='UTC')
-  ))
+  try_formats = c("%Y-%m-%d %H:%M:%OS", "%Y/%m/%d %H:%M:%OS", "%Y-%m-%d %H:%M", "%Y/%m/%d %H:%M", "%Y-%m-%d", "%Y/%m/%d")
+  dateformat <- try_formats[sapply(try_formats, function(fmt) all(!is.na(as.POSIXct(unlist(times), format=fmt))))]
```
lindsaycarr: is this related to fixing the issue with `c()` or a new feature??

aappling-usgs: Related. Rather than trying to get rid of the `c()`, which was introducing new headaches, I decided it was simpler to apply the c() to strings rather than POSIXct objects, and then parse them back into POSIXct after the concatenation. But to do that I needed to have the format written out, and I know we discussed in slack that it might be fine to assume the format of the dates is always '%Y-%m-%d', but by the time we had had that conversation I had figured out this approach which would allow us to specify hours if we wanted. In the end, if the format of `dates` is always like `2018-08-23`, then `dateformat` should always equal `%Y-%m-%d`.

aappling-usgs:
So my Harvey test is complete, at least through the fetching and processing of the precip data. Strange thing, though: I'm not actually seeing any difference in the resulting precip rasters, or even in 1_fetch/out/precip_data.rds.ind, which has the exact same hash it had before I rebuilt. And the names of the precip_rasters also look fine to me: they range from 2017-08-25 00:00:00 UTC through 2017-09-01 23:00:00 UTC with exactly 1 hour between each raster, no more or less. So...success? Or more strangeness?

lindsaycarr:
Well, that is 1 hour before the actual date entered right? `2017-09-02` is the "end time" in the config.  So, if "2017-09-02 00:00:00" is a timestep, then this would eventually fail. At least that's what is happening in mine.

aappling-usgs:
Oh! Is it just failing on that 1 timestep?

lindsaycarr:
well, it was more than that because of the timezone thing but yeah, I feel like it would still be one off right?

aappling-usgs:
:woman_facepalming:  it sure seemed that we had identified a real bug before, which should have caused us to be missing several hours on the tail end, but then (1) the "fix" didn't change anything, which i'm finding bewildering, and (2) yeah, you're right, this fencepost issue (missing the end point) is also a problem.

i do know how to fix the 2nd problem and will push a change in a sec.

aappling-usgs:
Harvey too!

PR #136, opened 2018-09-11
DON'T MERGE!!! Forecast Precip

dblodgett-usgs:
I still need to test that this doesn't break any of the old stageiv functionality, but figured I'd get the guts of it opened in a PR now that it's ready.

### DIFF
--- a/1_fetch/src/fetch_precip_nc.R
+++ b/1_fetch/src/fetch_precip_nc.R
@@ -3,6 +3,38 @@
 #' @param times the start and end time of the request
 fetch_precip_data <- function(ind_file, view_polygon, times) {
 
+  now <- Sys.time()
+  attributes(now)$tzone <- "UTC"
+  start <- as.POSIXct(times$start, tz = "UTC", format = "%Y-%m-%d")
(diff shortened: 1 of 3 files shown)


dblodgett-usgs:
forecast_prcp

1_fetch/src/fetch_precip_nc.R
```diff
@@ -3,6 +3,38 @@
 #' @param times the start and end time of the request
 fetch_precip_data <- function(ind_file, view_polygon, times) {
 
+  now <- Sys.time()
+  attributes(now)$tzone <- "UTC"
+  start <- as.POSIXct(times$start, tz = "UTC", format = "%Y-%m-%d")
+  end <- as.POSIXct(times$end, tz = "UTC", format = "%Y-%m-%d")
+  if(start  > now) {
```
aappling-usgs: what if start < now but end > now?

aappling-usgs: Would it be possible to include both historical and forecast precip in the same vizzie? I can imagine that being really compelling partway through a storm

PR #138, opened 2018-09-11
fix to allow specific times in viz_config start/end dates

lindsaycarr:
Needs #138 to actually work.

### DIFF
--- a/1_fetch/src/fetch_precip_nc.R
+++ b/1_fetch/src/fetch_precip_nc.R
@@ -25,6 +25,10 @@ fetch_precip_data <- function(ind_file, view_polygon, times) {
   # user's local timezone
   try_formats <- c("%Y-%m-%d %H:%M:%OS", "%Y/%m/%d %H:%M:%OS", "%Y-%m-%d %H:%M", "%Y/%m/%d %H:%M", "%Y-%m-%d", "%Y/%m/%d")
   dateformat <- try_formats[sapply(try_formats, function(fmt) all(!is.na(as.POSIXct(unlist(times), format=fmt))))]
+  # choose only the first that matches because technically "2018-09-11 00:00:00"
+  # matches "%Y-%m-%d %H:%M:%OS", "%Y-%m-%d %H:%M", and "%Y-%m-%d"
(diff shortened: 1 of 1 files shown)


1_fetch/src/fetch_precip_nc.R
```diff
@@ -25,6 +25,10 @@ fetch_precip_data <- function(ind_file, view_polygon, times) {
   # user's local timezone
   try_formats <- c("%Y-%m-%d %H:%M:%OS", "%Y/%m/%d %H:%M:%OS", "%Y-%m-%d %H:%M", "%Y/%m/%d %H:%M", "%Y-%m-%d", "%Y/%m/%d")
   dateformat <- try_formats[sapply(try_formats, function(fmt) all(!is.na(as.POSIXct(unlist(times), format=fmt))))]
+  # choose only the first that matches because technically "2018-09-11 00:00:00"
+  # matches "%Y-%m-%d %H:%M:%OS", "%Y-%m-%d %H:%M", and "%Y-%m-%d"
+  # so we just need it to choose the first of these, otherwise it causes an error
+  dateformat <- dateformat[1]
```
jread-usgs: better to just delete lines 26 & 27 and use `dateformat <- "%Y-%m-%d %H:%M:%OS"` to avoid confusion in the code?

lindsaycarr: @aappling-usgs and I have these to give flexibility for the dates. At least for specifying time or having just year-month-day. I can see an argument for not supporting the flexibility. We could just all agree that we need to put the hours:minutes:seconds in there.

aappling-usgs: deleting lines 26 and 27 wouldn't be the same - note the filtering of try_formats in line 27. (we could move the `[1]` to occur at the end of line 27 instead of in its own line here, but that's a style question)

agreed that somewhat less flexibility would be fine, though i think we might as well support some option for specifying times because it's basically in there at this point. i could see having just one option for specifying dates and one for specifying date-times. the try_dates code i added is just pulled straight from as.POSIXct and so mirrors the default options allowed by that function.

PR #139, opened 2018-09-12
move precip spatial and values to process

dblodgett-usgs:
This PR splits the precip processing up so we pull the nc in fetch, save it, then process it twice -- once to create the spatial content, once to create the values. All the processing is now moved to the process phase and three separate artifacts get saved. 

I also removed the iteration on the nc data download since our firewall issue has been resolved.

I'm going to merge these changes into florence locally and see what I see with the precip next. A future PR to Florence will likely include this commit.


dblodgett-usgs:
These changes are included in my PR to florence #141. If we want to close that PR, I can rework my PR so it's only changes to master and update this.

dblodgett-usgs:
Tested this with Harvey precip. Looking good.
screen shot 2018-09-13 at 8 11 41 am

PR #141, opened 2018-09-13
Precip data pull fix.

dblodgett-usgs:
This PR fixes the precipitation pull issue I debugged last night. I have not rebuilt the shared project with this code but have tested the pull to work.
We were getting: 
screen shot 2018-09-12 at 9 00 45 pm

And now we get:
screen shot 2018-09-13 at 5 57 14 am


1_fetch/src/fetch_precip_nc.R
```diff
@@ -2,67 +2,60 @@
...
-  stencil_pts <- as.stencil_pt('xmin', 'ymin', 'low_left') %>% cbind(stencil_pts)
+  
+  precip_boundary <- "POLYGON ((-80.74672 19.80368, -119.023 23.117, -134.043 53.51192, -124.9736 55.90588, -114.5051 57.4236, -102.7265 57.82357, -91.28471 56.95466, -81.85043 55.19577, -73.63054 52.69872, -66.4331 49.53428, -59.95137 45.6187, -80.74672 19.80368))" %>%
+    st_as_sfc(crs = st_crs("+proj=longlat +datum=WGS84"))
+  
+  stencil <- st_transform(view_polygon, "+proj=longlat +datum=WGS84") %>%
+    st_intersection(precip_boundary)
+  
+  stencil <- as_Spatial(stencil)
+  stencil@proj4string <- sp::CRS("+proj=longlat +datum=WGS84") # geoknife complains otherwise
```
lindsaycarr: so we need `sp` objects for geoknife?

viz_config.yml
```diff
@@ -7,7 +7,7 @@ date_display_tz: 'UTC'
 
 # view, dimensions
 bbox: [-83, 25, -67, 39]
-projection: "+proj=lcc +lat_1=36.16666666666666 +lat_2=34.33333333333334 +lat_0=33.75 +lon_0=-79 +x_0=609601.2192024384 +y_0=0 +ellps=GRS80 +datum=NAD83 +to_meter=0.3048006096012192 +no_defs"
+projection: "+init=epsg:2264"
```
lindsaycarr: can you explain why we need to switch the projection?

1_fetch/src/fetch_precip_nc.R
```diff
@@ -2,67 +2,60 @@
...
-    st_sf(st_sfc(st_point(c(bbox[[box1]], bbox[[box2]]), dim = "XY")), crs = crs) %>%
-      st_transform(crs = "+init=epsg:4326") %>%
-      st_coordinates() %>% as.vector() %>% data.frame(x = .) %>% setNames(col_name)
-  }
-  stencil_pts <- as.stencil_pt('xmin', 'ymax', 'up_left')
-  stencil_pts <- as.stencil_pt('xmax', 'ymax', 'up_right') %>% cbind(stencil_pts)
-  stencil_pts <- as.stencil_pt('xmax', 'ymin', 'low_right') %>% cbind(stencil_pts)
-  stencil_pts <- as.stencil_pt('xmin', 'ymin', 'low_left') %>% cbind(stencil_pts)
+  
+  precip_boundary <- "POLYGON ((-80.74672 19.80368, -119.023 23.117, -134.043 53.51192, -124.9736 55.90588, -114.5051 57.4236, -102.7265 57.82357, -91.28471 56.95466, -81.85043 55.19577, -73.63054 52.69872, -66.4331 49.53428, -59.95137 45.6187, -80.74672 19.80368))" %>%
```
lindsaycarr: so this is unique to the Florence branch. Any reason we shouldn't have this in the main code?

1_fetch/src/fetch_precip_nc.R
dblodgett-usgs: The code that was used to create this was subject to axis order issues. Safer to stick to sf and sp standard methods than try to mess with coordinates. This change isn't critical but is safer IMHO.

viz_config.yml
dblodgett-usgs: Didn't change, just using the epsg code rather than the fill proj4 string. I was worried that something about the proj parameters was causing the problem so I used the code to hit the database.

1_fetch/src/fetch_precip_nc.R
dblodgett-usgs: No. This is the boundary of the precip data source (stageiv). This will be the same for all.

dblodgett-usgs:
Closing the PR against this branch. Realizing that this is really better to go to master.

PR #143, opened 2018-09-13
Florence 2018

jread-usgs:


jread-usgs:
new colors and different timing

Issue #156, opened 2018-09-15
right-size GIFs for insta/twitter/facebook/reddit


jread-usgs:

jread-usgs:
We'll need to first investigate size/aspect recommendations. I think we can do our landscape view for all but insta (which is typically square), but the file size cutoffs for twitter are a little limiting, because we'll want to be below 5MB. Perhaps we build all GIFs below this cutoff, or we build the twitter GIF by skipping every other timestep's png frame when the GIF is built to keep it under 5MB. 

We'll have an early decision-point as to whether we'll build a special GIF for insta that isn't in landscape mode. If we do, we might want to use `dry_put = TRUE` and branch from Florence-2018.

jread-usgs:
looks like we'll want 1080 x 1080 for instagram

jread-usgs:
Facebook would support 1200 x 630 if we want to snag that extra bit of resolution vs twitter (1024 x 512)

PR #162, opened 2018-09-15
Florence 2018 instagram

jread-usgs:


jread-usgs:
Part of #156

6_visualize/src/prep_spark_line_fun.R
```diff
@@ -98,7 +98,7 @@ prep_spark_funs_data <- function(stage_data, site_data, timestep_ind, spark_conf
         data_frame(dateTime = head(flood_stage_line$dateTime, 1), stage_normalized = flood_stage_va),
         flood_stage_line %>% select(dateTime, stage_normalized),
         data.frame(dateTime = tail(flood_stage_line$dateTime, 1), stage_normalized = flood_stage_va))
-
+
```
aappling-usgs: What do you have your whitespace handling set to?

vizstorm-GIF-Florence-2018.Rproj
```diff
@@ -11,7 +11,3 @@ Encoding: UTF-8
 
 RnwWeave: knitr
 LaTeX: pdfLaTeX
-
-AutoAppendNewline: Yes
-StripTrailingWhitespace: Yes
```
aappling-usgs: oh, this is why the trailing whitespace is not being stripped

aappling-usgs: not a huge deal, but i do like how setting this to Yes reduces the number of whitespace changes we make in PRs

jread-usgs: oh, didn't realize this. this must be the default. I started a new project w/ an empty dir, and then pointed the remotes to github to pull and build from there. I probably should've kept the original .proj that got pulled down.

PR #167, opened 2018-09-15
RDG outro frame

aappling-usgs:
gif_frame_a_outro_rdgs_1

My plan is to manually add targets to 6_outro_gif_frames.yml: the MVP would be text+points for the long-term network in one frame that precedes this RDG frame, text for the URL in one frame that follows this RDG frame. 

I put a placeholder in there for specifying the alpha, which would allow us to make multiple RDG frames with varying alphas to show it fading in. But alpha doesn't actually do anything yet.


6_visualize/src/prep_outro_funs.R
```diff
@@ -0,0 +1,47 @@
+prep_outro_rdgs_fun <- function(rdg_ind="1_fetch/out/rapid_dep_sites.rds.ind", gage_col_config, outro_placement, legend_text_cfg, opacity=1) {
+  # project from lat/lon to plot coordinates. . this properly belongs in a 2_process step, but we're releasing tomorrow
+  proj_str <- "+proj=lcc +lat_1=34.83333333333334 +lat_2=32.5 +lat_0=31.83333333333333 +lon_0=-81 +x_0=609600 +y_0=0 +ellps=GRS80 +units=m +no_defs "
```
jread-usgs: hard-coding for now 👍

aappling-usgs: yep. deserves a fix later if we keep this outro

PR #173, opened 2018-09-16
Florence 2018

jread-usgs:
I put a good bit of hack hard coding for the outro frames because we are releasing today. I was hoping to wait for the next round of precip data to update...but that hasn't happened yet.


6_visualize/src/prep_outro_funs.R
```diff
@@ -1,96 +1,119 @@
...
+  allsites <- readRDS(sc_retrieve(allsites_ind))
+
+  if(opacity != 1) stop("opacity other than 1 not yet supported")
+
+  ltn_pch <- 19
+  ltn_col <- gage_col_config$gage_ltn_col
+  ltn_cex <- 1
+
+  font_add_google('Abel', "abel")
+  font_add_google('Oswald', "Oswald")
```
aappling-usgs: nice

PR #177, opened 2018-09-19
09-19 build

jread-usgs:


2_process/src/clean_streamdata.R
```diff
@@ -1,18 +1,18 @@
...
-      NA,
-      stage_normalized)
-  )
-  
+
+  # cleaned <- mutate(
+  #   normalized,
+  #   stage_normalized = ifelse(
+  #     (site_no == '02091500' & dateTime >= as.POSIXct('2018-09-15 08:00:00', tz='UTC')) |
+  #       (site_no == '02092500' & dateTime >= as.POSIXct('2018-09-15 17:00:00', tz='UTC')),
```
lindsaycarr: so we don't find these two points suspicious anymore?

jread-usgs: they were removed

PR #181, opened 2018-10-09
core changes made for Florence

aappling-usgs:
I compared the Florence-2018 branch to the master branch and pulled in all changes except the additions of non-example files, .ind files, and build/status/*.yml files. I think this is probably good to go, but another set of eyes doing a quick skim would be useful to make sure I didn't add in anything that would get in the way of future vizzies.


aappling-usgs:
Jordan votes for Insta portrait dimensions as our one and only format. Instagram max portrait size is 1080px x 1350px, but 720 width x 900 height worked well for Florence (and produces smaller / quicker-loading files)

PR #192, opened 2018-10-13
missed indicator files

wdwatkins:


aappling-usgs:
Closing because we determined these three .ind files have already been overwritten

PR #203, opened 2019-03-22
snow raster interpolation function

wdwatkins:
I've not built this into the scipiper parts yet, but this should work for either one large interpolation or doing separate ones for separate files.  This function takes a timestep size and an arbitrary number of daily files and interpolates to that timestep.  It just needs the read/write parts filled in a bit , which will depend on how what pattern we use with scipiper.  

The interpolation part took ~8-10 minutes to run interpolating between three rasters.  It is basically doing a separate interpolation for each raster cell, so is crunching a lot of numbers.  I suspect doing one big interpolation would be faster, although you could conceivably parallelize doing  chunks of days at a time.  

I commented in several options for writing out files -- could make a raster timeseries, or just subset back into separate layers and write out a file for each timestep.  I'm not sure how that would work with scipiper though -- it wouldn't be a 1:1 input/output relationship, is that handled yet?


2_process/src/process_snow_rasters.R
```diff
@@ -1,41 +1,73 @@
...
+  steps <- seq(from = 0, by = timestep_in_hours,
+               length.out = (hours_covered/timestep_in_hours)+1)
+  have_data <- which(steps %% 24 == 0) #these are the steps to interpolate between
+  #insert all NA layers, then na.spline interpolates NA values
+  empty <- rep(NA, length(steps))
+  fun <- function(y) {
+    if(all(is.na(y))) {
+      empty
+    } else {
+      zoo::na.approx((empty[have_data] <- y), xout=1:length(steps))
```
wdwatkins: need to change NAs to zeros or else this errors.  `na.approx` does linear interpolation, `na.spline` does splines

lindsaycarr: I changed to zeros in #204 (https://github.com/USGS-VIZLAB/vizstorm-GIF/commit/8646034a574443c10fb21e29002dab85cef47241)

2_process/src/process_snow_rasters.R
```diff
@@ -1,41 +1,73 @@
...
   data_file <- as_data_file(ind_file)
   saveRDS(snow_raster_proj_crop, data_file)
   gd_put(remote_ind=ind_file, local_source=data_file, mock_get='none')
-  
+
+}
+
+interpolate_snow_raster_layers <- function(timestep_in_hours, ...) {
+
+  snow_raster_rds_inds <- c(...) #would be either passing in dates or files here?
```
lindsaycarr: I think you could pass in `'1_fetch/log/1_fetch_snow_tasks.ind'` which looks like this:

```
2_process/out/raster_snow_20190312.rds.ind: ccbaa2f343760d20d1a846a5b57049a2
2_process/out/raster_snow_20190313.rds.ind: 3f8ca1de5d72d6ce7ef525813c3c21cd
2_process/out/raster_snow_20190314.rds.ind: ce9e9b7eb900a7668d7aa96286c5e2a8
2_process/out/raster_snow_20190315.rds.ind: a5e779e55d1d1c3a4d4d2d589a8f552d
2_process/out/raster_snow_20190316.rds.ind: 822d8c7e2fa0ce8753f7fdbd45c2ac58
2_process/out/raster_snow_20190317.rds.ind: 0ef163be6308485bbbf9e77da02aa52b
2_process/out/raster_snow_20190318.rds.ind: d6a56c66f4d819c0642e0927aef09dc8
2_process/out/raster_snow_20190319.rds.ind: a315d56862bb6bee432275567abfd587
```

and you can get all the filenames by reading it in as yaml:

```
snow_raster_rds_inds <- names(yaml::yaml.load_file('1_fetch/log/1_fetch_snow_tasks.ind'))
snow_raster_rds_inds

[1] "2_process/out/raster_snow_20190312.rds.ind" "2_process/out/raster_snow_20190313.rds.ind"
[3] "2_process/out/raster_snow_20190314.rds.ind" "2_process/out/raster_snow_20190315.rds.ind"
[5] "2_process/out/raster_snow_20190316.rds.ind" "2_process/out/raster_snow_20190317.rds.ind"
[7] "2_process/out/raster_snow_20190318.rds.ind" "2_process/out/raster_snow_20190319.rds.ind"
```

lindsaycarr:
> it wouldn't be a 1:1 input/output relationship, is that handled yet?

I think that is what @jread-usgs branch for `task_combiners` in `scipiper` is all about (https://github.com/USGS-R/scipiper/tree/task_combiners), but I know it's not in the master branch yet.

We could have this process step for interpolating live outside of the snow tasks and create an rds list with everything. Passing around that list could be expensive, I'm honestly not sure. I had originally implemented it that way, but that was before I cropped things. Although, now we are adding a bunch of timesteps. If that is still expense, we could add another step to `1_fetch_snow_tasks.yml` that takes the rds list of everything, subsets to timesteps within one day, and then saves those as individual files. Then, the visualize functions will be handling slightly smaller files and can subset to the specific timestep in their functions. 

My proposal is coming as a PR to your branch.

wdwatkins:
Hmm, we would need to be able to distinguish them from actual zeros

wdwatkins:
And we would need to set certain interpolated cells to NA as well

lindsaycarr:
If it is 0 (no snow), then I think we want it to be NA. So if you're converting them from 0 to NA after the interpolation, that should work right?

wdwatkins:
Ok, if zero and NA are interchangeable, then yeah it's easy

wdwatkins:
Wasn't sure if needed to distinguish between actual missing data and 0

lindsaycarr:
Nope, I don't think that is necessary

PR #204, opened 2019-03-22
Midwest winter flooding 2019 - variety of updates, including gages & snow raster

lindsaycarr:
- mp4 instead of gif
- state outlines on top of snow
- shift view slightly more south
- change custom gage filter (keep all gages that will flood in states along the MS river, keep 100 random sites in states along MS, keep another 50 random sites in view)
- slight style changes to water mark and timestamp
- change bad snow values to 0 instead of NA (for @wdwatkins interpolation work)


2_process/src/process_snow_rasters.R
```diff
@@ -16,7 +16,7 @@ process_snow_raster <- function(ind_file, snow_data_ind, snow_data_yml, crop_ext
   
   snow_depth <- readBin(snow_data_fn, integer(), 
                         n=n_row*n_col, size=2, signed=TRUE, endian='big')
-  snow_depth[snow_depth <= 0] <- as.numeric(NA)
+  snow_depth[snow_depth <= 0] <- 0
```
jread-usgs: :+1:

Issue #207, opened 2019-03-23
interpolate to sub-daily snowpack


jread-usgs:
We are working on this for #205 and having some trouble using functions on the rasters themselves. As a quick example, we can do these operations quickly by using the raw files (as vectors) before converting to rasters. 

Assuming we have the daily files locally, we can generate interpolated values for each hour between them by weighting the time distance between day 0 and day 1. To show this, here is an example of building hourlies from two files. I'm choosing a single location (index that has snow to show that the operation happens to each index and is linear between the two actual (daily) values:

```r

n_col <- 8192
n_row <- 4096
snow_depth0 <- readBin('~/Downloads/SNODAS_unmasked_20190319/zz_ssmv11036tS__T0001TTNATS2019031905HP001.dat', 
                      integer(), n=n_row*n_col, size=2, signed=TRUE, endian='big')
snow_depth0[snow_depth0 <= 0] <- 0
snow_depth1 <- readBin('~/Downloads/SNODAS_unmasked_20190320/zz_ssmv11036tS__T0001TTNATS2019032005HP001.dat', 
                       integer(), n=n_row*n_col, size=2, signed=TRUE, endian='big')
snow_depth1[snow_depth1 <= 0] <- 0

plot(c(0, 24), c(snow_depth0[9000000], snow_depth1[9000000]), col = 'red', pch = 16, ylab = 'depth', xlab = 'hours since day 0')
for (hour in 1:23){
  snow_depth_interp <- (snow_depth0 * (1 / hour) + snow_depth1 * (1 / (24-hour))) / (( 1 / hour) + (1 / (24-hour)))
  points(hour, snow_depth_interp[9000000], pch = 'x', col = 'green')
}
```

![image](https://user-images.githubusercontent.com/2349007/54866224-3d67c500-4d3f-11e9-8a64-b84f2e720512.png)

Assuming you have the files locally, the interpolation takes 5~10 seconds for 23 hours, which seems reasonable. 

to write to rasters, 
```r
snow_raster <- raster::raster(ncol=n_col, nrow=n_row, xmn = x0, xmx = x1, ymn = y0, ymx = y1)
snow_raster <- raster::setValues(snow_raster, snow_depth_interp)
```

PR #208, opened 2019-04-01
Midwest winter flooding 2019 - get David's changes into plotting code

lindsaycarr:
David added an attempt to get interpolation working. This works those interpolated rasters into the plotting code. Changes were actually made 3/22, but I am just now committing. Resulting animation had some weird choppiness going on, so we need to look for a different interpolation method.


lindsaycarr:
Committing these changes in preparation to try out a different interpolation method (see #207)

PR #209, opened 2019-04-01
Midwest winter flooding 2019 - new interpolating method

### LINKED ISSUE #207: interpolate to sub-daily snowpack

lindsayplatt:
Fixes #207. 

This creates a raster stack with a raster for every timestep. This does not get pushed up to Google Drive because the transfer takes way longer than building it locally does. In the future, we can look at creating a task table to create a file for every timestep.

I also changed the date times for UTC so that the Central times would start at 2019-03-12 00:00:00 rather than starting at 2019-03-11 19:00:00. This caused almost everything to rebuild which took ~ 1 hr.


1_fetch.yml
```diff
@@ -192,12 +194,12 @@ targets:
   # -- get additional gage sites for outtro -- #
   stn_event:
     command: viz_config[[I('stn_event')]]
-  1_fetch/out/rapid_dep_sites.rds.ind:
-    command: fetch_rapid_dep_sites(
-      ind_file = target_name,
-      event = stn_event)
-  1_fetch/out/rapid_dep_sites.rds:
-    command: gd_get('1_fetch/out/rapid_dep_sites.rds.ind')
+  #1_fetch/out/rapid_dep_sites.rds.ind:
```
jread-usgs: I'd suggest delete instead of commenting it out

2_process/src/process_snow_rasters.R
```diff
@@ -40,51 +40,66 @@ process_snow_raster <- function(ind_file, snow_data_ind, snow_data_yml, crop_ext
...
-
+  
+  raster_stack <- raster::stack(x = rasters_interp_list)
+  
   data_file <- as_data_file(ind_file)
-  saveRDS(interp_raster_stack, data_file)
-  gd_put(remote_ind = ind_file)
+  raster::writeRaster(raster_stack, data_file, overwrite=TRUE)
+  
+  # It is quicker to build locally, then push + pull a file of this size (~700 MB for 8 days)
```
jread-usgs: agreed!

6_visualize.yml
```diff
@@ -32,10 +32,8 @@ targets:
   6_visualize:
     depends:
       - 6_storm_gif_tasks.yml
-      - 6_visualize/log/6_storm_gif_tasks.ind
-      - 6_intro_gif_tasks.yml
-      - 6_visualize/log/6_intro_gif_tasks.ind
-      - 6_visualize/out/animation_a.gif
+      - storm_gif_tasks_executed
```
jread-usgs: since `6_visualize/out/animation_a.mp4` depends on `storm_gif_tasks_executed`, you shouldn't need to include it here too (right?)

lindsaycarr:
There's still a bit of weirdness between end of Mar 11 & beginning of Mar 12 as well as end of Mar 18 and beginning of Mar 19. Could be something that didn't get rebuilt because those line up with the dates of the previous window.

ezgif com-video-to-gif

6_visualize.yml
lindsaycarr: Oh oops, you're right. Added it at the top first because I was trying to get it to run when I just did `scmake('6_visualize/out/animation_a.mp4', '6_visualize.yml', force=TRUE)` and remembered the `depends` trick for individual targets after. Removing!

PR #210, opened 2019-05-23
Update data through 5/20

lindsayplatt:
Refreshed data (all fetch and process steps) to be up-to-date through 5/20/2019.

`streamdata.rds.ind` kept rebuilding every time. Couldn't figure out why, but visualize steps depend on a processed step that ingests streamdata, so we should be good now.

Ran: 
```{r}
library(scipiper)
scmake(remake_file = "1_fetch.yml")
scmake(remake_file = "2_process.yml")
```


lindsayplatt:
Merging on my own since they are extremely small changes. Also, want the code `ind` to reflect what is on Google Drive.

PR #213, opened 2019-05-29
Non-GDP precip pull

wdwatkins:
The precip spatial processing step after the pull worked, but haven't test this any further down the chain


wdwatkins:
This just computes the indices to grab from THREDDS in R, and then runs a system `nccopy` command to get the actual netcdf file.  Theoretically this should give us the same result as the GDP's OPeNDAP subset algorithm 🤞 .

dblodgett-usgs:
Nice work! I totally spaced on the fact that you need to pull down the NetCDF file! 

I implemented this: https://github.com/USGS-R/intersectr/pull/16 to help with that! Although I have not tested it on 2d axes. Looks like what you did here basically emulates that function.

PR #215, opened 2019-05-30
Final adjustments to Central US Spring 2019 flooding event gif

lindsayplatt:
Struggled a lot with getting frame delays to work how I would expect.


lindsayplatt:
Merging since this work was done over a year ago.

PR #218, opened 2020-06-24
Leftover fixes from previous build

lindsayplatt:
Includes some older fixes that should be on master (including fixes so that the number of rapid deploy gages can be 0 & using task combiners at the end of at task makefile to use as input to the combine_animation_frames fxn) + some changes to the storm used in the viz_config.yml file which should be just fine.


6_visualize/src/create_gif_makefile.R
```diff
@@ -10,7 +10,7 @@ create_intro_gif_makefile <- function(makefile, task_plan, remake_file) {
     file_extensions=c('feather','ind'))
 }
 
-create_storm_gif_makefile <- function(makefile, task_plan, remake_file) {
+create_storm_gif_makefile <- function(makefile, task_plan, remake_file, final_targets) {
```
aappling-usgs, line 13: The `final_targets`argument doesn't appear to be used in this call to `create_task_makefile`, but that's where it should appear (along with a value for `finalize_funs`) if any combining is going to happen...do you think this may be a case where we started to try to add a combiner and then aborted partway through?

6_visualize/src/prep_outro_funs.R
```diff
@@ -1,7 +1,15 @@
 prep_outro_rdgs_fun <- function(rdg_ind="1_fetch/out/rapid_dep_sites.rds.ind", gage_col_config, outro_placement, legend_text_cfg, opacity=1) {
   # project from lat/lon to plot coordinates. . this properly belongs in a 2_process step, but we're releasing tomorrow
   proj_str <- "+proj=lcc +lat_1=34.83333333333334 +lat_2=32.5 +lat_0=31.83333333333333 +lon_0=-81 +x_0=609600 +y_0=0 +ellps=GRS80 +units=m +no_defs "
-  rdgs <- readRDS(sc_retrieve(rdg_ind)) %>%
+
+  rdgs <- readRDS(sc_retrieve(rdg_ind))
```
aappling-usgs, line 5: New pattern we just started trying in `delaware-model-prep` is to put all getter targets in a separate, disconnected remake file called "getters.yml". For now, at least, that would also mean adding a `remake_file='getters.yml'` argument to every call to `sc_retrieve` throughout the project. (But if you wanted to start by making a PR in scipiper to add a default `remake_file='getters.yml'` to `sc_retrieve` you could save some typing in this repo).
  This may not be a change you want to make in this PR, but it has saved us numerous unnecessary rebuilds in that other project and I think it would be a noticeable improvement for vizstorm projects, too.

6_visualize/src/create_gif_makefile.R
aappling-usgs: If I'm remembering right, these code changes may have been the point at which you created this scipiper issue https://github.com/USGS-R/scipiper/issues/116.

6_outro_gif_frames.yml
```diff
@@ -22,14 +22,14 @@ targets:
       view_fun,
       basemap_fun,
       ocean_name_fun,
-      precip_raster_fun_a_20180919_09,
+      precip_raster_fun_a_20190715_12,
```
aappling-usgs, line 25: There are a bunch of little changes likes these in the PR that I'd have to run code to evaluate. Have you run this code this spring, Lindsay, such that you're at least confident that these style/data changes don't break the pipeline or ruin the gif?

README.md
```diff
@@ -1,5 +1,9 @@
 # vizstorm-GIF
 
+# This needs scipiper task combiners
+
+This branch requires scipiper with task combiners!
```
aappling-usgs, line 5: This should be a less important warning now that we've merged combiners into the main scipiper...but may as well leave it here.

viz_config-example.yml
```diff
@@ -105,82 +104,20 @@ cities:
...
-    #  text_cex: 0.8
+      name: "Houston TX"
     -
-      name: "New Bern NC"
-      news: "CBS"
-      dot_cex: 0.8
-      text_cex: 0.8
-      lat: 35.107686
-      long: -77.051366
-    -
```
aappling-usgs, line 142: I don't remember how we're managing multiple hurricanes within this repo - different branches, maybe? - but It might be nice to keep some of these customization examples in here if this is the template for future hurricane vizzies.

6_outro_gif_frames.yml
lindsayplatt: TBH, I'm not really sure where some of these changes came from (they are things that should only be in the `Makerspace-example` branch, which picked a smaller storm with no hurricane. I thought it wouldn't be the end of the world to just leave them rather than revert, but some of your comments below make me want to revert those. 

Those changes are mostly reverting https://github.com/USGS-VIZLAB/vizstorm-GIF/pull/218/commits/376ae2e03c649e7491d5a7aed704fc65ed8bf7da and https://github.com/USGS-VIZLAB/vizstorm-GIF/pull/218/commits/ab2af0a5e264ef9a0d587c5b5a2ea4ac25635ae7 and the change this comment is referring to.

6_visualize/src/prep_outro_funs.R
lindsayplatt: Could you explain what you mean by `getter` targets? If this is something that has proven useful to stop rebuilds in other vizstorm projects, then I think it makes sense here. We have plans to do some overhaul on this pipeline but that isn't the highest priority, so I think it would be best to capture these ideas in an issue and revisit later.

6_visualize/src/create_gif_makefile.R
lindsayplatt: You are right, that is exactly what happened. I got part way and then had to figure out a different approach. This relates to our pipeline overhaul plan which is something we want to do soon, but not an immediate priority. I will delete this so that there isn't confusion.

README.md
lindsayplatt: Although, likely not even true since we aren't actually using the `final_targets` in the make file.

6_visualize/src/prep_outro_funs.R
aappling-usgs: Oh, sorry, I did mean to elaborate on that "getter" thing - I meant targets whose command is `gd_get`, `s3_get`, etc. See https://github.com/USGS-R/scipiper/issues/70#issuecomment-643671102. The requisite overhaul for implementing this involves changing a lot of files, but it's pretty much just a search-and-edit operation, so it was refreshingly fast to do for `delaware-model-prep`.

lindsayplatt:
@aappling-usgs I got rid of almost everything except the rapid deployment gages stuff 👍

PR #219, opened 2020-06-25
Get old pipeline building - new NLDI service & yml linkages

lindsayplatt:
I updated to scipiper to 0.0.20 (I forget what version this repo was last built with), but it was giving a lot of implicit target errors which I fixed by using `remake.yml` to link all of the other makefiles. In addition, NLDI changed from cida.usgs.gov to labs.waterdata.usgs.gov and I fixed that (though, it does have intermittent failures but just rebuilding eventually gets you passed those). Lastly, I was struggling with geoknife and precip data coming back for the wrong dates. This appears to be an issue with GDP and the specific dates I was using, so I am ignoring for now and just skipping over precip to get this to build (though, the skips aren't committed here).

Update: I did run this using `viz_config-example.yml` as my `viz_config.yml` which is using Hurricane Florence dates rather than Hurricane Barry (from July 2019 and in the `Makerspace-example` branch) and the precip data from `geoknife` downloaded as expected.


lindsayplatt:
I also just added the creation of the outro makefile as part of the pipeline instead of standalone and committed so that you didn't have to remember to update it yourself.

6_visualize/src/visualize_utils.R
```diff
@@ -11,3 +11,13 @@ existing_png_tasks <- function(path='6_visualize/tmp', pattern='a_[[:digit:]]{8}
   tasks <- regmatches(files, regexpr(sprintf("(%s)", pattern), files))
   return(tasks)
 }
+
+timesteps_to_build <- function(timestep_ind, frame_step = 1) {
```
jread-usgs, line 15: :+1:

6_visualize.yml
```diff
@@ -176,16 +180,15 @@ targets:
...
-      folders = visualize_folders,
-      frame_step = storm_frame_step)
+      folders = visualize_folders)
 
   6_storm_gif_tasks.yml:
     command: create_storm_gif_makefile(
       makefile = target_name,
       task_plan = storm_gif_tasks,
-      remake_file = I('6_visualize.yml'))
+      remake_file = I('remake.yml'))
```
jread-usgs, line 191: :+1:

6_visualize.yml
```diff
@@ -35,6 +34,7 @@ targets:
       - 6_visualize/log/6_storm_gif_tasks.ind
       - 6_intro_gif_tasks.yml
       - 6_visualize/log/6_intro_gif_tasks.ind
+      - 6_visualize/log/6_outro_gif_tasks.ind
       - 6_visualize/out/animation_a.gif
```
jread-usgs, line 38: is there a reason the `6_visualize/out/animation_a.gif` isn't the only depends here? Since the others go into the animation target as inputs

lindsayplatt: No reason that I know of. That can certainly be changed. This is likely just a leftover practice from the original implementation of this pipeline back when I didn't understand them completely :) I'd love to do an overhaul after I finish up Pipelines III ... so many improvements can be made!

PR #220, opened 2020-06-29
Add a timeline and replace the timestamp

lindsayplatt:
Instead of having the timestamp just displayed as text, this shows it as a timeline aligned with the sparklines. It helps give a visual picture of where you are in the animation + adds a bit more context for the sparklines. Removed the precip raster when I was building this because GDP wasn't working and built using the updates in #219. Please ignore the styling in the outro frames - I didn't commit changes there.

animation_timeline


6_visualize/src/create_gif_makefile.R
```diff
@@ -17,10 +17,11 @@ create_storm_gif_makefile <- function(makefile, task_plan, remake_file) {
     packages=c('dplyr', 'scipiper'),
     sources=c(
       '6_visualize/src/create_animation_frame.R',
-      '6_visualize/src/prep_datetime_fun.R',
       '6_visualize/src/prep_gage_sites_fun.R',
       '6_visualize/src/prep_storm_point_fun.R',
       '6_visualize/src/prep_precip_fun.R',
-      '6_visualize/src/prep_spark_line_fun.R'),
+      '6_visualize/src/prep_spark_line_fun.R',
```
jread-usgs, line 23: one thing we learned in pipelines III is that we've got some reproducibility issues with running the task tables unless we use `sources` as a depends or as inputs to the task table combiner function/output. 

We should consider if we want to make minimal edits to how this pipeline works for pragmatic reasons, or update it with the preferred task table structure.

6_visualize/src/prep_timeline_fun.R
```diff
@@ -0,0 +1,59 @@
+prep_timeline_fun <- function(DateTime, timestep_ind, timeline_config, spark_config, legend_text_cfg) {
+
+  this_DateTime <- as.POSIXct(DateTime, tz = "UTC") # WARNING, IF WE EVER MOVE FROM UTC elsewhere, this will be fragile/bad.
```
jread-usgs, line 3: are we displaying the UTC time or the local timezone time? Seems like the former, but don't viewers of the viz want to see the local tz?

6_visualize/src/prep_timeline_fun.R
```diff
@@ -0,0 +1,59 @@
...
+
+  # Compute the full x limits for all datetimes
+  storm_timesteps <- fetch_read(timestep_ind)
```
jread-usgs, line 6: seems like a lot of reading (and fetching?) this tiny file when an arg to `timeline_fun_a_20180912_06` could be `date_lims = I(c('2018-01-01','2018-01-03'))`

6_visualize/src/create_gif_tasks.R
```diff
@@ -182,15 +182,21 @@ create_storm_gif_tasks <- function(timestep_ind, storm_track_cfg, folders, frame
...
-      sprintf('datetime_fun_%s', task_name)
+      sprintf('timeline_fun_%s', task_name)
     },
     command = function(task_name, ...){
       cur_task <- dplyr::filter(rename(tasks, tn=task_name), tn==task_name)
-      sprintf("prep_datetime_fun(I('%s'), datetime_placement, date_display_tz)", format(cur_task$timestep, "%Y-%m-%d %H:%M:%S"))
+      psprintf(
+        "prep_timeline_fun(",
+        "DateTime = I('%s'),"=cur_task$timestep,
+        "timestep_ind = '2_process/out/timesteps.rds.ind',",
```
jread-usgs, line 196: see my note below on replacing `timestep_ind` with a date_lim arg that is calculated in the task_plot builder step.

6_visualize/src/prep_timeline_fun.R
```diff
@@ -0,0 +1,59 @@
...
+    loc_date_y <- coord_space[3] + spark_config$ytimeline_text * diff(coord_space[3:4])
+
+    # Convert this the current date & start/end to user coordinates
+    loc_date_x <- dateTime_to_x(this_DateTime, loc_x)
+    loc_start_x <- dateTime_to_x(date_lims[1], loc_x)
+    loc_end_x <- dateTime_to_x(date_lims[2], loc_x)
+
+    # Add timeline
+    points(loc_x, rep(loc_y, length(loc_x)), type = 'l', col = timeline_config$line_col,
+           lwd = timeline_config$line_lwd, lty = timeline_config$line_lty)
```
jread-usgs, line 36: I think a tiny vertical tick for the marker may be a better choice than the triangle, since the triangle adds another "marker" (e.g., dots, circles) to the mix and you can avoid that w/ using a small tick instead? just a thought. I think you were looking at different options here.

viz_config-example.yml
```diff
@@ -35,6 +35,16 @@ legend_text:
   cex: 1.5
   col: "#161616"
   family: ""
+timeline_style:
+  pt_cex: 1.5
+  pt_col: "gray40"
+  pt_pch: 25
```
jread-usgs, line 41: see comment on marker type above.

6_visualize/src/prep_timeline_fun.R
lindsayplatt: Good question. When [we showed the timestamp, we displayed that in the local timezone](https://github.com/USGS-VIZLAB/vizstorm-GIF/blob/master/6_visualize/src/prep_datetime_fun.R#L4) but UTC is what is being passed around throughout all of the frame creation. It would make sense to convert the dates to local time before displaying here, too.

6_visualize/src/prep_timeline_fun.R
lindsayplatt: The vertical bar feels a bit dated to me. I sketched out a few different ideas [here in Teams](https://teams.microsoft.com/l/message/19:79eaaf0209574972b1afd7c5453fc11e@thread.skype/1593439144163?tenantId=0693b5ba-4b18-4d7b-9341-f32f400a5494&groupId=80983861-ab36-4fdc-a217-17742ba5ee3d&parentMessageId=1593188119326&teamName=GS-WMA-IIDD-Makerspace&channelName=General&createdTime=1593439144163) which is how the triangle idea came to be. 

Here's what it looks like with a bar:

animation_a

lindsayplatt:
In response to your pipeline brittleness comments with task tables - I am treating that as a completely separate effort and would like to focus this PR review on just the timeline and pipeline pieces that are needed for the timeline.

6_visualize/src/prep_timeline_fun.R
jread-usgs: yeah, that looks dated :). Was thinking more like this:
image (or even smaller)

lindsayplatt: OK, so that is definitely better than the tick version I implemented. I am still on the fence though. I like the simplicity but I'm struggling to think about how modern or not modern it is. Probably better than an arrow, though.

I'm not sure I would want it even smaller because then it might not be seen at all.

lindsayplatt: Most recent commit results in:

animation_a

jread-usgs: what do you think after seeing it? Personally, I prefer it because it is simple and doesn't introduce a new marker type, but I wouldn't shed tears over you going in the other direction :)

lindsayplatt: I do like the simplicity and the kind of "wiping" effect it creates with moving at the same time as the spark lines. I also liked the triangle but don't feel like it makes anything that much better so I am good with going for simple!

jread-usgs: :shipit:

lindsayplatt:
Final version after review & reconciliation: 

animation_a

PR #221, opened 2020-07-06
Update the hurricane path to be a comet

lindsayplatt:
Previous critiques of the old path was that it looked like a crack in someone's screen, especially on mobile devices. This adjusts the storm path to be dynamic for each timestep (which makes it look like it has a comet tail) rather than a solid line for the whole animation. 

This PR also adjusts the Google font issue (where you forget to load it in a new session, build all your frames with a lame font, and then go "oh crap") by loading the font at session startup through a `.Rprofile` file.

Updated ([see the "cracked screen" version here](https://github.com/USGS-VIZLAB/vizstorm-GIF/pull/220#issue-441547949)):

animation_a


.Rprofile
```diff
@@ -0,0 +1,5 @@
+# Load fonts used at the start of every session, otherwise it won't be able to use them
+# Will need to change if the fonts use change
+
+message("Loading font 'Abel' from .Rprofile")
+sysfonts::font_add_google('Abel', 'abel')
```
jread-usgs, line 5: :+1:

6_visualize/in/6_outro_gif_tasks_template.mustache
```diff
@@ -23,7 +23,7 @@ targets:
       basemap_fun,
       ocean_name_fun,
       precip_raster_fun_a_{{last_task}},
-      storm_line_fun,
+      storm_line_fun_a_{{last_task}},
```
jread-usgs, line 26: wow, have we always used mustache templating for outro tasks? seems fine to me (and I like using whisker/mustache), just wondering why the deviation from a task table? This question is out of scope for this particular review, but I am curious nonetheless.

6_visualize/src/prep_storm_line_fun.R
```diff
@@ -1,12 +1,70 @@
...
+        # Plot each fading section individually
+        # For the last one, use the last row in the data
+        n_start <- nrow(before_this_dot) - tail_lengths[i]
+        n_end <- nrow(before_this_dot) - ifelse(i==length(tail_lengths), 0, tail_lengths[i+1])
+
+        # For the initial ones where tail_lengths may be > nrow
+        if(n_start < 1) { n_start <- 1 }
+        if(n_end < 1) { n_end <- nrow(before_this_dot) }
+
+        plot(st_geometry(slice(before_this_dot, n_start:n_end)), add=TRUE,
```
jread-usgs, line 36: huh, cool way to do this :+1:

viz_config-example.yml
```diff
@@ -46,6 +48,12 @@ timeline_style:
   font_cex: 1
   font_col: "gray30"
 
+storm_line:
```
jread-usgs, line 51: this works, but wondering why `storm_line_col` isn't within the `storm_line:` list. 

Also, not really encouraging a change here, but wondering how you decide which style elements end up in the yaml vs those that end up in the function. Such as the `lty`.

6_visualize/src/prep_storm_line_fun.R
```diff
@@ -1,12 +1,70 @@
...
+             col=cols[i], type = 'l', lty="dotted", lwd = 2)
+      }
     }
   }
   return(plot_fun)
 }
+
+# source("6_visualize/src/visualize_utils.R")
```
jread-usgs, line 44: Do you want to delete this or put it in the README?

6_visualize/in/6_outro_gif_tasks_template.mustache
lindsayplatt: This came along with the previous PR (#220) when I noticed we just had this file manually built. I didn't think this was a great candidate for task tables because none of the targets repeat. There are really only 3 output targets (3 frames) and they are all different, but share the value stored in `last_task`. Although, maybe I was approaching it wrong. I kept thinking that the "tasks" are the frame dates, but that is what worked in the storm gif tasks recipe. Maybe the "tasks" here are `rdgs`, `ltn`, and `readmore`.

6_visualize/src/prep_storm_line_fun.R
lindsayplatt: Oops - nope, that shouldn't be there.

6_visualize/in/6_outro_gif_tasks_template.mustache
jread-usgs: got it, good to know. I think I worked on these frames in the past but don't remember how we stitched them together.

viz_config-example.yml
lindsayplatt: TBH, it is kind of arbitrary right now. I've been hard-coding things in the function unless I thought that I wanted to actively change it. 

And as for where this `storm_line_col` styling went ... the legend function gets passed everything under `# styling` already and this is the color that determines what the line looks like in the legend. Rather than adding an additional argument to the legend fxn, I just added it there. 

The viz_config could probably use some cleanup anyways .... if we group into `hurricane_style` rather than just have all of the prefixed `hurricane_` style cfgs, it would probably be less daunting to just add every possible style cfg, like `lty`, to the `viz_config` because it would be more organized.

lindsayplatt:
I have tested that functionality before (in [this branch](https://github.com/USGS-VIZLAB/vizstorm-GIF/tree/MidwestWinterFlooding-2019)), but not with these new changes. Although, now after looking at that, I think there are some differences in `1_fetch.yml`. I'll add some changes to handle that case in the master branch.

lindsayplatt:
OK, so pulled over the required changes and when you comment out `storm_code` and `stn_event` in `viz_config.yml`, the storm line, storm point, and subsequent legend entries are removed.

With `storm_code` and `stn_event`:

image


Without `storm_code` and `stn_event`:

image

jread-usgs:
nice!

lindsayplatt:
are we calling this 👍 for now?

jread-usgs:
I think so, yes

PR #224, opened 2020-09-16
shift bounding box east and north

collnell:
Modifies bounding box to be centered on Mobile, MS where Sally landed, and include FL panhandle, AL, GA, SC, and NC where flooding from rain is expected.

gif_TEST_frame_a_20200915_12


lindsayplatt:
Also changes end time to be closer to current time.

PR #226, opened 2020-09-16
Sally 2020 - larger stream orders only + precip working

lindsayplatt:
Removed some of the stream detail. Accepted that we probably won't try to find a new data source for waterbodies for this viz and will just leave out that detail. Also, there was a services issue for the precip data on GDP, but it has been resolved and we have usable precip data now.


2_process/src/process_precip.R
```diff
@@ -141,7 +141,7 @@ process_precip_rasters <- function(ind_file, precip_spatial_ind,
   precip_spatial <- readRDS(sc_retrieve(precip_spatial_ind)) %>%
     dplyr::select(-x, -y)
 
-  time <- dplyr::select(precip, time_chr) %>%
+  time_df <- dplyr::select(precip, time_chr) %>%
```
jread-usgs, line 144: weird, was this a bug before? I don't see how it would've worked w/o this edit

viz_config.yml
```diff
@@ -211,7 +211,7 @@ cities:
     - "Jacksonville FL"
 
 # Only streamorders greater than this will be fetched.
```
jread-usgs, line 213: greater _or equal_ to this will be fetched.

jread-usgs (approved):

2_process/src/process_precip.R
lindsayplatt: It was a part I forgot to copy from Laura. It failed before it actually tried to use `time_df` because of the data only having one timestep this time around, so I only noticed it once we got past that

PR #227, opened 2020-09-17
Sally 2020 - build full animation

lindsayplatt:
OK, so this is mostly working! There are some obvious placement issues and a need to tweak gages shown + cities shown, but it gets us pretty close. The only issue was that I changed the date to something that I think was too close to real time when I initiated the fetch and so it didn't pull down data for the last 2 frames and caused in error in trying to plot NULL data. I did some manual work so that I could attached a GIF to this PR and did'nt have time to change the date and rebuild everything. So, someone will need to repull the data and build all of that before the full pipeline will run from end-to-end again. I think the key is to pick times that are > 6 hrs in the past.

Current* state:

animation_a


6_visualize/src/prep_precip_fun.R
```diff
@@ -15,7 +15,8 @@ prep_precip_fun <- function(precip_rasters, precip_bins, timestep){
   rm(precip_rasters, precip_bins, timestep)
 
   plot_fun <- function(){
+    par(mai = c(0,0,0,0))
```
jread-usgs, line 18: do you know why this was needed?

2_process/src/filter_sites_custom.R
```diff
@@ -2,16 +2,46 @@
...
   # get sites data frame with NWS data
   sites_df <- readRDS(sc_retrieve(sites_ind))
 
   ### filter sites_df here ###
   set.seed(303)
 
-  sites_info_subset <- sites_df %>%
-    sample_n(6)
+  # Get sites in desired states
+  sites_state_specific <- dataRetrieval::readNWISsite(sites_df$site_no) %>%
```
jread-usgs, line 14: we will update this w/ hand-chosen sites. Placeholder for now :+1:

jread-usgs (approved):

PR #228, opened 2020-09-18
Sally 2020 gage location additions

jread-usgs:
This is odd and I don't have time this morning to figure this out: I am trying to simply define the sites we're using, but for some reason the sparklines and gage locations aren't propagating to the .png build

I am going to leave this open in case someone can take a look


jread-usgs:
Fixed now with https://github.com/USGS-VIZLAB/vizstorm-GIF/pull/228/commits/6c71f6bd5a52b4a815b478bef0eeda4488de394d so this can be merged as a starting point for hand-selected sites.

Issue #239, opened 2021-08-24
Fixes during `Henri-2021` that need to get migrated to `main`


lindsayplatt:
Running list of everything we had to adjust that needs to make it into the `main` repo:

- [ ] Switch river geom fetch to using nhdplustools as the source (see https://github.com/USGS-VIZLAB/vizstorm-GIF/commit/4bd309a21050592e931473800d69e8dff9a86cb2#diff-1aa42baff6599120ec10682d10f92b046bacb4237e17359cba55397949618b0b)
- [ ] Implement the `sp` stencil bug issue mentioned in #237 (see https://github.com/USGS-VIZLAB/vizstorm-GIF/blob/Henri-2021/1_fetch/src/fetch_precip_nc.R#L24)
- [ ] Add the `remake_file = getOption()` solution since this pipeline was built before the switch to the `getters.yml` approach. Although, might be able to set the option for the whole pipeline instead of adding this in a bunch of places (see https://github.com/USGS-VIZLAB/vizstorm-GIF/commit/7ac3ed7c236cc7bddb492834abf0647bb2494073#diff-1965d7a54599d150f5f6e5609f4db440418fee35d77dcdb3e3e1460a4dcf18be)
- [ ] Strip out the `dim()` attribute before binding columns (see https://github.com/USGS-VIZLAB/vizstorm-GIF/commit/a007951bcfcf1bc815f1708986a640e08c93b4fe)
- [ ] Add timestep names back to the precip rasters (see https://github.com/USGS-VIZLAB/vizstorm-GIF/commit/1330bc67b1cfc61144dc8057f52f16430c17903e)
- [ ] Include the `par(mai = c(0,0,0,0))` for the precip plot function bc we discovered that the setting was somehow getting changed during Sally (see https://github.com/USGS-VIZLAB/vizstorm-GIF/commit/bbec9f1887da2865394761831cda8bdb90442209)
- [ ] Edit the mustache template for the outro gif tasks to use `timeline_fun_a` and not `datetime_fun_a`. We never made that switch there after we implemented the new time visual solution (see ???)
- [ ] Add note to `viz_config.yml` about how to look up `stn` code (see https://github.com/USGS-VIZLAB/vizstorm-GIF/blob/Sally-2020/viz_config.yml#L4)

Look at the final files in the `Henri-2021` branch that go with the commits above in case there were additional commits that improved those. Just included those to help orient you to the needed changes.

jread-usgs:
[...]

PR #241, opened 2021-08-28
fixes from `Henri-2021` needed for build

collnell:
Addresses most issues noted in https://github.com/USGS-VIZLAB/vizstorm-GIF/issues/239:
- Uses `nhdplusTools` to fetch river and waterbody geoms
- Adds sp stencil bug fix
- Adds `stn_Event` link to config and remove duplicate `storm_line_col`
- Removes dim to allow precip bind by date
- Directs `sc_retrieve` to remake over default to `getters.yml`
- Updates outro frames to use `timeline_fun` added during `Sally-2020`
- Adds `par` to plotting function for `precip_rasters` to correct misalignment with river geoms

### DIFF
--- a/1_fetch/src/fetch_river_geometry.R
+++ b/1_fetch/src/fetch_river_geometry.R
@@ -3,11 +3,8 @@ fetch_major_river_geoms <- function(ind_file, view_polygon, fetch_streamorder) {
   bbox_sf <- st_as_sfc(st_bbox(view_polygon))
 
   sf_major_rivers <- get_nhdplus(AOI = bbox_sf,
-                                 realization = "flowline",
-                                 streamorder = fetch_streamorder$fetch_streamorder) %>%
-    group_by(gnis_id, streamorde) %>%
-    summarize() %>%
-    ms_simplify()
+                           realization = "flowline",
+                           streamorder = fetch_streamorder$fetch_streamorder)
 
   data_file <- as_data_file(ind_file)
   saveRDS(sf_major_rivers, data_file)
@@ -19,7 +16,7 @@ fetch_waterbody_geoms <- function(ind_file, view_polygon, fetch_waterbody_areasq
   bbox_sf <- st_as_sfc(st_bbox(view_polygon))
 
   sf_waterbodies <- get_waterbodies(AOI = bbox_sf,
-                                    buffer = 0) %>%
+                                 buffer = 0) %>%
     filter(areasqkm > fetch_waterbody_areasqkm$fetch_waterbody_areasqkm) %>%
     ms_simplify()
 
@@ -30,7 +27,7 @@ fetch_waterbody_geoms <- function(ind_file, view_polygon, fetch_waterbody_areasq
 
 fetch_gage_river_geoms <- function(ind_file, view_polygon, sites_ind) {
 
-  sites <- readRDS(sc_retrieve(sites_ind))
+  sites <- readRDS(sc_retrieve(sites_ind, remake_file = getOption("scipiper.remake_file")))
 
   if(nrow(sites) > 30) {
     set.seed(42)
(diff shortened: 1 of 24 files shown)


1_fetch/src/fetch_river_geometry.R
```diff
@@ -101,9 +54,12 @@ fetch_gage_river_geoms <- function(ind_file, view_polygon, sites_ind) {
 }
 
 name_adder <- function(x, updn) {
+  browser()
   lapply(seq_along(x), function(y, n, r, updn) {
+    print(y)
```
lindsayplatt, line 59: We should delete these lines leftover from troubleshooting

1_fetch/src/fetch_sites.R
```diff
@@ -71,7 +71,7 @@ fetch_rapid_dep_sites <- function(ind_file, event) {
     rdgs <- data.frame()
   } else {
     # fetch the rapid deployment gages (RDGs) from the STN web services
-    url <- sprintf('https://stn.wim.usgs.gov/STNServices/Instruments/FilteredInstruments.json?Event=%d&EventType=&EventStatus=&States=&County=&CurrentStatus=&CollectionCondition=&SensorType=5&DeploymentType=', event)
+    url <- sprintf('https://stn.wim.usgs.gov/STNServices/Instruments/FilteredInstruments.json?Event=%d&EventType=&EventStatus=&States=&County=&CurrentStatus=&CollectionCondition=&SensorType=1&DeploymentType=1,2', event)
```
lindsayplatt, line 74: I don't think we want to commit this change - this is what allowed us to show sensors other than the "rapid deployment gages", but I think we should make that decision to expand beyond RDGs per event.

2_process/src/process_precip.R
```diff
@@ -190,6 +194,11 @@ process_precip_rasters <- function(ind_file, precip_spatial_ind,
 
   parallel::stopCluster(cl)
 
+  # Name the rasters the same as the timestep they represent
+  #   8.23.2021: this part used to just happen and downstream
+  #   steps depend on the naming to be like this.
+  names(rasters) <- as.character(time$time)
```
lindsayplatt, line 200: So this makes it seem like the variable in this object is `time` but line 188 above makes it seem like it is `date_time`. I would expect them to use the same one?

2_process/src/process_river_geoms.R
```diff
@@ -1,16 +1,15 @@
 process_river_geoms <- function(ind_file,
                                 major_river_geoms_ind,
-                                gage_river_geoms_ind,
```
lindsayplatt, line 3: We don't want to commit any changes related to skipped `gage_river_geoms` - these were special for getting Henri out the door. We want to keep the `main` branch clean.

6_visualize/in/6_outro_gif_tasks_template.mustache
```diff
@@ -29,7 +29,7 @@ targets:
       gage_sites_fun_a_{{last_task}},
       rdgs_fun_1,
       legend_fun_a_{{last_task}},
-      datetime_fun_a_{{last_task}},
+      timeline_fun_a_{{last_task}},
```
lindsayplatt, line 32: I think we actually should just remove these from the outro entirely, as we did in Henri. In the outro, it would be frozen on the same date anyways ...

viz_config-example.yml
```diff
@@ -32,7 +32,6 @@ gage_river_col: "#5d9ddc"
 lake_col: "#a2c7eb"
 marsh_col: NA
 ocean_col: "#e7f1fa"
-storm_line_col: "grey70"
```
lindsayplatt, line 35: oh! I didn't notice there was a duplicate 👏

lindsayplatt (changes requested):
Just a few things I don't think we want in this PR. Otherwise 👍

collnell:
Feedback has been addressed.

PR #243, opened 2021-09-14
Add target for creating a video & fix delays in the gif

### LINKED ISSUE #242: Add video target
In addition to the GIF, include a target that creates a video as requested via Twitter. Consider making this one that file that is uploaded to Drupal for people to download into the future. Can mostly borrow [the function from gage-conditions-gif](https://github.com/USGS-VIZLAB/gage-conditions-gif/blob/master/6_visualize/src/combine_animation_frames.R#L1-L22).

lindsayplatt:
Fixes #242 by adding an additional target to `6_visualize.yml` - the mp4 file. In addition, I was noticing some funkiness in the two recent GIFs with how the final frames were timed (they didn't have the same, steady delay). I figured out it was due to an issue with NAs and fixed this for future GIF animations.

### DIFF
--- a/6_visualize/src/combine_animation_frames.R
+++ b/6_visualize/src/combine_animation_frames.R
@@ -1,4 +1,50 @@
-combine_animation_frames <- function(gif_file, animation_cfg, frame_ind_intro=NULL, frame_ind_storm=NULL, frame_ind_outro=NULL, intro_config) {
+combine_animation_frames_video <- function(out_file, animation_cfg, frame_ind_intro=NULL, frame_ind_storm=NULL, frame_ind_outro=NULL, intro_config) {
+
+  # Use ffmpeg to build a video
+
+  # Gather appropriate file names (not just whatever is in the directory)
+  png_files_intro <- extract_filenames_from_ind(frame_ind_intro)
+  png_files_storm <- extract_filenames_from_ind(frame_ind_storm)
+  png_files_outro <- extract_filenames_from_ind(frame_ind_outro)
+
+  # Create separate videos for intro, storm, and outro
+  make_video <- function(video_out, png_files, frame_delay) {
+
+    # FFMPEG only works on files numbered sequentially, so rename temporarily
+    file_name_df <- tibble(origName = png_files,
+                           countFormatted = zeroPad(1:length(png_files), padTo = 3),
+                           newName = file.path("6_visualize/tmp", paste0("frame_", countFormatted, ".png")))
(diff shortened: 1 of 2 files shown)


6_visualize/src/combine_animation_frames.R
```diff
@@ -33,24 +79,28 @@ combine_animation_frames <- function(gif_file, animation_cfg, frame_ind_intro=NU
...
   # freeze the last storm frame too for as long as we are showing each outro frame:
-  last_storm_delay <- calc_delays(freeze_delay, total_frames-n_outro, total_frames-n_outro)
-  outro_delays <- calc_delays(outro_delay, total_frames-n_outro+1, total_frames-1)
-  final_delay <- calc_delays(final_delay, total_frames, total_frames)
+  last_storm_delay_str <- calc_delays(freeze_delay, total_frames-n_outro, total_frames-n_outro)
+  outro_delay_str <- calc_delays(outro_delay, total_frames-n_outro+1, total_frames-1)
+  final_delay_str <- calc_delays(final_delay, total_frames, total_frames)
 
-  gifsicle_command <- sprintf('gifsicle -b -O3 %s %s %s %s %s %s --colors 256', gif_file, intro_delays, storm_delays, last_storm_delay, outro_delays, final_delay)
+  gifsicle_command <- sprintf('gifsicle -b -O3 %s %s %s %s %s %s --colors 256', gif_file,
```
collnell, line 95: Do you know any reference that explains gifsicle commands?

collnell: I haven't been able to

collnell (approved):
Looks goooood. 

Thinking about what you said about adding the black bar at the bottom for the video. I am guessing it would overlap plot items that fall in the lower space? What if there were different dimensions for that version, or are there reasons to not do that?

6_visualize/src/combine_animation_frames.R
jread-usgs: https://www.lcdf.org/gifsicle/man.html useful?

lindsayplatt: ^ that and stackoverflow are where I find most of my gifsicle help!

lindsayplatt:
@collnell I haven't tested here yet, but I add the bar to the gage-conditions animation by slightly scaling the existing video after it is created. I think that can work here, too. See [this code](https://github.com/USGS-VIZLAB/gage-conditions-gif#to-create-a-usgs-visid-compliant-video-version). I didn't get to that the other day!

PR #244, opened 2021-09-17
Adjust how logos are added + include visid banner version

lindsayplatt:
We needed to create a version with the VisID black bar and USGS logo on the bottom. To do so, we needed to create frames that did not already have the logo on them so that it wasn't duplicated (couldn't just cover up the logo as we do in gage-conditions because it isn't over a solid background). 

Thus, this PR:

1. Removes the use `prep_watermark_fun()` for each each frame.
2. Creates a `_nologo` version of both the GIF and video files
3. Adds the grey logo to the topleft for both the GIF and video files
4. Uses the `_nologo` video file and appends the visid black bar at the bottom (this creates a slightly different size video fiile)

Currently having this build all the way through in Henri as an example to make sure it works. Will un-draft it when that is complete and add some screenshots, so you can see the differences.

Gif/video without logo:
image

Gif/video with logo:
image

Video with VisID banner:

image


lindsayplatt:
Might want to pick an `in/` logo that I don't need to make transparent or remove the background. Seems to be missing parts. Would be simple enough to just replace `6_visualize/in/usgs_logo_black.png`. 

image

ceenell (approved):
Looks good once that logo is swapped out. Thanks for the quick turnaround on this

lindsayplatt:
New logo in video: 

image

New logo in GIF (still a bit grainy... :( but looks good in video):

image

lindsayplatt:
Since there is still some graininess, we may just want to release the video file on twitter, too. I don't think we see the autoplay issues with videos that we used to.

